PR #14, opened 2017-01-01
use pouchdb-http package instead of pouchdb

nolanlawson:
Since you're only using the PouchDB HTTP adapter and the `.changes()` API, you can get away with just using the `pouchdb-http` package, which only contains `pouchdb-core` and `pouchdb-adapter-http`. Notably it avoids installing LevelDB so will lead to faster npm/yarn installs: http://pouchdb.com/custom.html


vvo:
Hi @nolanlawson thanks a lot for this!

PR #15, opened 2017-01-02
Tweak the searchableAttributes settings to consider popularName first.

redox:
This will require a full re-indexing :)

### DIFF
--- a/config.js
+++ b/config.js
@@ -10,6 +10,7 @@ const defaultConfig = {
   seq: null,
   indexSettings: {
     searchableAttributes: [
+      'unordered(popularName)',
       'unordered(name)',
       'unordered(description)',
       'unordered(keywords)',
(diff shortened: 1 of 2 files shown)


vvo:
Do you have an example bad query that will be good afterwards?

vvo:
> This will require a full re-indexing :)

IZY!

redox:
> Do you have an example bad query that will be good afterwards?

Yes, `asyc` for instance.

vvo:
deploying and doing full reindex

vvo:
Was done but result is maybe not what we want:

https://yarnpkg.com/search?q=asyc&p=0&is_v=1

2017-01-03-103911_1380x958_scrot

redox:
Yes, I spot that; I need to iterate on the ranking formula now.

PR #21, opened 2017-03-01
feat(dependents): add dependents to the index

Haroenv:
> note: not used in the process yet

another option is using https://github.com/nice-registry/dependent-packages/


Haroenv:
closed until it's finished

vvo:
Can you include only the relevant parts without syntax modification/prettification?

Maybe do #25 and then reformat the whole code so we do not have anymore this.

Haroenv:
yep of course, just opened because I found it weird to work on a closed branch

Haroenv:
switched to npms. Can not be merged until we have confirmation it's fine to use their api.

iam4x:
@Haroenv It feels weird to use `npms.io` API since they are trying to build better search like us?

config.js
```diff
@@ -19,7 +19,7 @@ const defaultConfig = {
       'owners.name',
     ],
     attributesForFaceting: ['onlyFilter(name)', 'keywords'], // optionalFacetFilters to boost the name
-    customRanking: ['desc(downloadsLast30Days)'],
+    customRanking: ['desc(downloadsLast30Days)', 'desc(dependents)'],
```
vvo: This will have no impact because desc(downloadsLast30Days) is already too precise so dependents would only be taken into account if downloads are the same.

What we could do is have a custom ranking that is made of silos (like group 1,2,3,4,5,6,7..) and then use dependents.

If I am wrong sorry I do not remember exactly what downloadsLast30Days includes.

Haroenv: It's not exactly the same, but something with a lot of dependents will also get downloaded more

PR #24, opened 2017-03-09
feat(data): Sanitize html in readmes

Haroenv:
npm provides this, and in case their readme is not available (happens sometimes), we can then still get the readme from github as a fallback

### DIFF
--- a/formatPkg.js
+++ b/formatPkg.js
@@ -53,6 +53,7 @@ export default function formatPkg(pkg) {
     githubRepo,
     gitHead,
     readme: pkg.readme,
+    readmeFilename: pkg.readmeFilename,
     owner,
     deprecated: cleaned.deprecated !== undefined ? cleaned.deprecated : false,
     homepage: getHomePage(cleaned.homepage, cleaned.repository),
--- a/formatPkg.js
+++ b/formatPkg.js
@@ -1,4 +1,4 @@
-import URL from 'url';
+import {URL} from 'url';
 import NicePackage from 'nice-package';
 import gravatarUrl from 'gravatar-url';
 import numeral from 'numeral';
--- a/formatPkg.js
+++ b/formatPkg.js
@@ -1,8 +1,11 @@
+import URL from 'url';
 import NicePackage from 'nice-package';
 import gravatarUrl from 'gravatar-url';
 import numeral from 'numeral';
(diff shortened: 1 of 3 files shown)


Haroenv:
We should resolve from GitHub in this case

vvo:
Yes, resolve most of the data backend side when feasible (i.e. no rate limits involved) so that you do not have to do much work frontend wise.

package.json
```diff
@@ -27,11 +27,14 @@
...
     "gravatar-url": "^1.1.0",
     "lodash": "^4.17.4",
+    "marked": "^0.3.6",
     "ms": "^0.7.2",
     "nice-package": "^3.0.1",
     "numeral": "^2.0.4",
     "pouchdb-http": "^6.0.2",
-    "traverse": "^0.6.6"
+    "traverse": "^0.6.6",
+    "url": "https://github.com/webcomponents/URL",
```
vvo: Seems like you can use url-polyfill: https://github.com/webcomponents/URL/blob/812ca9b2baae653290aceb36362d219cb96c0f35/package.json

I try to avoid dependencies as http/s/git, it's usually slow to install and has no explicit versioning.

Haroenv: nope, it's not published on npm, should I publish it myself?

formatPkg.js
```diff
@@ -68,10 +75,58 @@ export default function formatPkg(pkg) {
...
+      this.update(escape(node));
+    }
   }
 }
 
+const prefixURL = (url, { base, user, project, head, path }) => {
+  if (url.indexOf('//') > 0) {
+    return url;
+  } else {
+    return url
```
vvo: double return, maybe time to put eslint

yarn.lock
```diff
@@ -3803,6 +3807,10 @@ url@^0.11.0:
     punycode "1.3.2"
     querystring "0.2.0"
 
+"url@https://github.com/webcomponents/URL":
```
vvo: To completely remove a module with yarn you need to use yarn remove name-of-module/url I guess

Haroenv: fixed by regenerating lockfile

vvo:
For reference, the default xss whitelist is: https://github.com/leizongmin/js-xss/blob/84d06c74bab55c885e35453a688ef49545389588/lib/default.js#L12

PR #26, opened 2017-03-10
fix(pkg.readme) don't render markdown

Haroenv:
We decided to save bandwidth and place in the Algolia index by storing the raw markdown instead of the rendered and escaped html. The readme isn't been run through `escape-html`, since then we can't render the markdown properly on the client

### DIFF
--- a/formatPkg.js
+++ b/formatPkg.js
@@ -4,8 +4,6 @@ import gravatarUrl from 'gravatar-url';
 import numeral from 'numeral';
 const defaultGravatar = 'https://www.gravatar.com/avatar/';
 import escape from 'escape-html';
-import marked from 'marked';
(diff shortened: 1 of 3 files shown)


formatPkg.js
```diff
@@ -45,7 +43,7 @@ export default function formatPkg(pkg) {
 
   const _readme = pkg.readme;
   // todo: fetch from github if _readme is undefined
-  const readme = _readme && html({ markdown:_readme, githubRepo, gitHead });
+  const readme = _readme;
```
vvo: We do not anymore needs this temp variable I guess

Haroenv: nope, but I left it here, in case we move the fetching from GitHub here

PR #27, opened 2017-03-10
feat(pkg.readme) truncate readmes longer than 450kB

Haroenv:
This won't happen often at all, since for now, it should never appear. 

adds ` **TRUNCATED**` to the end of a very long README.

### DIFF
--- a/formatPkg.js
+++ b/formatPkg.js
@@ -1,10 +1,12 @@
-import {URL} from 'url';
+import sizeof from 'object-sizeof';
 import NicePackage from 'nice-package';
 import gravatarUrl from 'gravatar-url';
 import numeral from 'numeral';
 const defaultGravatar = 'https://www.gravatar.com/avatar/';
 import escape from 'escape-html';
 import traverse from 'traverse';
+
+import c from './config';
 
 export default function formatPkg(pkg) {
   const cleaned = new NicePackage(pkg);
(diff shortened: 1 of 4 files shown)


formatPkg.js
```diff
@@ -1,11 +1,13 @@
-import {URL} from 'url';
```
Haroenv: this removal should actually have been done in the previous PR

formatPkg.js
```diff
@@ -45,7 +47,7 @@ export default function formatPkg(pkg) {
   // todo: fetch from github if _readme is undefined
   const readme = _readme;
 
-  return traverse({
+  const _package = {
```
Haroenv: I'd love to be able to call this `package`, but that's a reserved word

algobot: Try rawPackage or dirtyPackage since it's gonna be cleaned?

vvo: Try rawPackage or dirtyPackage since it's gonna be cleaned?

vvo:
Will this cut in the middle of `[link](http://...`?

Haroenv:
Yes, it will, but in that case the renderer will show

[link](https:// **TRUNCATED**

vvo:
Should we use a module that does this truncate? https://www.npmjs.com/package/truncate-utf8-bytes seems to fit our needs while removing custom code for us (no maintenance on this code)

Haroenv:
Good point, probably has very similar code

PR #28, opened 2017-03-12
feat(truncate) depend on truncate-utf8-bytes

Haroenv:
no need for custom code, and is actually technically better. See https://github.com/algolia/npm-search/pull/27#issuecomment-285950849

### DIFF
--- a/formatPkg.js
+++ b/formatPkg.js
@@ -30,7 +30,11 @@ export default function formatPkg(pkg) {
   }
 
   const version = cleaned.version ? cleaned.version : '0.0.0';
-  const gitHead = pkg.versions ? pkg.versions[version] ? pkg.versions[version].gitHead : undefined : undefined;
+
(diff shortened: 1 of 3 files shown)


formatPkg.js
```diff
@@ -5,6 +5,7 @@ import numeral from 'numeral';
 const defaultGravatar = 'https://www.gravatar.com/avatar/';
 import escape from 'escape-html';
 import traverse from 'traverse';
+import truncate from 'truncate-utf8-bytes';
```
Haroenv: Not really understanding why they loop through the whole content? Might be a problem for us?

https://github.com/parshap/truncate-utf8-bytes/blob/master/lib/truncate.js#L26-L39

vvo:
If we observe any slowdown (using our nodejs analyzer tool on heroku), we will just create a new package to truncate utf8 strings with benchmarks. Deal?

PR #30, opened 2017-03-15
feat(changelog) resolve changelog filename

Haroenv:
we do a promise.race-like thing, where errors are swallowed (promise.race returns either the first .catch or .then, see https://www.jcore.com/2016/12/18/promise-me-you-wont-use-promise-race/)


vvo:
Community FTW: https://github.com/Financial-Times/promise-rat-race :)

vvo:
(found by doing a search on an awesome npm search website.. :D)

Haroenv:
This takes about 150ms now (took 180ms with GET) for 4 packages to resolve.

Will depend on the package instead!

github.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  const {user, project, path} = githubRepo;
+  if (user.length < 1 || project.length < 1) {
+    return {changelogFilename: null};
+  }
+
+  const baseGithubURL = `https://raw.githubusercontent.com/${user}/${project}/${gitHead}/${`${path.replace('/tree/', '')}`}`;
+  const files = [
+    'CHANGELOG.md',
```
Haroenv: NPM uses a regex for this, but we obviously can't

https://github.com/npm/npm/blob/master/lib/utils/tar.js#L81

> `changes|changelog|history`

I don't check for `changes`/`changes.md`, but I think it's not often used.

Haroenv: looked it up, and that is used too. Added it

github.js
```diff
@@ -0,0 +1,50 @@
...
+    'CHANGES',
+    'changes',
+    'Changes',
+    'HISTORY.md',
+    'history.md',
+    'HISTORY',
+    'history',
+  ].map(file => [baseGithubURL.replace(/\/$/, ''), file].join('/'));
+
+  return race(files.map(got, {method: 'HEAD'}))
```
vvo: Nice! 💯

saveDocs.js
```diff
@@ -20,5 +21,13 @@ export default function saveDocs(docs) {
 }
 
 function addMetaData(pkgs) {
-  return npm.getDownloads(pkgs);
+  return Promise.all([npm.getDownloads(pkgs), getChangelogs(pkgs)]).then(([
```
vvo: 💥

PR #31, opened 2017-03-15
feat(deps) add dependencies

Haroenv:
easier than dependents because this is only one level deep.

### DIFF
--- a/formatPkg.js
+++ b/formatPkg.js
@@ -57,6 +57,8 @@ export default function formatPkg(pkg) {
     if (typeof cleaned.keywords === 'string') keywords = [cleaned.keywords];
   }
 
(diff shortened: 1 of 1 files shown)


redox:
Should we add `devDependencies` as well?

Haroenv:
We could in just the same way, but it's way less relevant for reusers of a package

redox:
> We could in just the same way, but it's way less relevant for reusers of a package

Sure, I think we rather need the opposite. If we now that a package X is always used as a dev dependencies; we'll be able to display `yarn add ... --dev` instead of `yarn add ....`.

Haroenv:
That’s true @redox, but would be more relevant for adding `devDependents` (but that’s in #21)

PR #34, opened 2017-03-28
feat(data): add concatenated name

### LINKED ISSUE #33: add alternative name without dashes
for `create-react-native-app` this should be `createreactnativeapp`. This is because it's often easier to just write it without dashes.

To remove is:

* `.`
* `-`
* `_`
* `@`
* `/`

The alternative name should have a lower relevancy score than the real name to avoid `react-native` to score above `reactnative` when you search for `reactnative`.

This should also be done for the keywords, but in that case the concatenated version can simply be added.

Haroenv:
fixes #33

### DIFF
--- a/formatPkg.js
+++ b/formatPkg.js
@@ -58,11 +58,12 @@ export default function formatPkg(pkg) {
   }
 
   const dependencies = cleaned.dependencies || {};
+  const concatenatedName = cleaned.name.replace(/[-/@_.]+/g, '');
 
   const rawPkg = {
     objectID: cleaned.name,
     name: cleaned.name,
-    numberOfWordsInName: cleaned.name.split(/[-/@_]+/).length,
+    concatenatedName,
     downloadsLast30Days: 0,
     downloadsRatio: 0,
     humanDownloadsLast30Days: numeral(0).format('0.[0]a'),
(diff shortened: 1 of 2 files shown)


formatPkg.js
```diff
@@ -58,11 +58,17 @@ export default function formatPkg(pkg) {
   }
 
   const dependencies = cleaned.dependencies || {};
+  const charsToSplit = /[-/@_.]+/;
```
Haroenv: added `.` to split on here too

formatPkg.js
```diff
@@ -58,11 +58,17 @@ export default function formatPkg(pkg) {
...
+    new RegExp(charsToSplit.source, 'g'),
+    '',
+  );
 
   const rawPkg = {
     objectID: cleaned.name,
     name: cleaned.name,
-    numberOfWordsInName: cleaned.name.split(/[-/@_]+/).length,
+    concatenatedName,
+    numberOfWordsInName: cleaned.name.split(charsToSplit).length,
```
iam4x:
```js
'r-a-b-c'.split(/[-/@_.]+/g)
-> (4) ["r", "a", "b", "c"]
```

Haroenv: thanks!

Haroenv:
I removed `numberOfWordsInName` because it's not useful anymore when the concatenated name strategy is used.

PR #39, opened 2017-04-29
fix(formatPkg): rewrite get info into separate functions

Haroenv:
The complexity of this function was too high according to eslint, so it wouldn't regularly let me commit. 

I broke all of the reassignments up into separate functions returning different versions.


vvo:
👍

PR #40, opened 2017-04-29
feat(process): redo bootstrap after X amount of time

### LINKED ISSUE #20: Automatically restart replication every week
To get a downloads data that is always up to date, we need to restart the replication from seq=0 every month.

There should be ways on heroku to do that.

Haroenv:
fixes #20

In the 'watch' step, I get the state after each iteration, then if now - last time bootstrap was done is more than the time to redo the bootstrap, the settings

```js
stateManager.set({
  seq: 0,
  bootstrapDone: false,
});
```

are being set, which *should* make the bootstrap start over.

I am not very sure how to test this locally without having to wait a long time...

### DIFF
--- a/index.js
+++ b/index.js
@@ -184,6 +184,9 @@ function watch({ seq }) {
         .then(({ bootstrapLastDone }) => {
           const now = Date.now();
           const lastBootstrapped = new Date(bootstrapLastDone);
+          // when the process is running longer than a certain time
+          // we want to start over and get all info again
+          // we do this by exiting and letting Heroku start over
           if (now - lastBootstrapped > c.timeToRedoBootstrap) {
             stateManager.set({
               seq: 0,
--- a/index.js
+++ b/index.js
@@ -104,7 +104,7 @@ function bootstrap(state) {
           log.info('⛷ Bootstrap: done');
           return stateManager.save({
             bootstrapDone: true,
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -179,6 +180,17 @@ function watch({ seq }) {
...
         )
+        .then(stateManager.get)
+        .then(({ bootstrapLastDone }) => {
+          const now = new Date();
+          const lastBootstrapped = new Date(bootstrapLastDone);
+          if (now - lastBootstrapped > c.timeToRedoBootstrap) {
+            stateManager.set({
+              seq: 0,
+              bootstrapDone: false,
+            });
```
Haroenv: I’m doubting if I need `.then(()=>exit(0))` is needed here.

redox: What do you think @vvo?

index.js
```diff
@@ -104,6 +104,7 @@ function bootstrap(state) {
           log.info('⛷ Bootstrap: done');
           return stateManager.save({
             bootstrapDone: true,
+            bootstrapLastDone: new Date().toISOString(),
```
vvo: Isn't Date.now() sufficient (store as a timestamp)?

index.js
vvo: Where would you want to put that exit(0)?

vvo: Where would you want to put that exit(0)?

PR #41, opened 2017-05-09
feat(pkg): add 'extra keywords' entry

Haroenv:
This is initially used to add a `yarn-create` keyword to packages that start with `create-` (like create-react-app, create-next-app, create-react-native-app etc.) These packages will be available on the yarn docs page about `yarn create`, so an extra keyword like this to filter is useful.

cc @cpojer

### DIFF
--- a/formatPkg.js
+++ b/formatPkg.js
@@ -159,15 +159,18 @@ function getVersions(cleaned) {
 }
 
 function getKeywords(cleaned) {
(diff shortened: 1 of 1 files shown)


Haroenv:
Since there will be *lots* and *lots* of packages that start with `create-` we decided that this isn’t useful for now, and rather we will add `yarn-create` keyword to the packages themselves for now.

PR #43, opened 2017-05-13
chore(trace): remove @risingstack/trace

Haroenv:
1. The trial expired
2. Unless we're on a pro plan, we can only see things like errors, which papertrail also did.
3. It always took very long to install, because it needed to compile native code

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,6 @@ import PouchDB from 'pouchdb-http';
 import * as npm from './npm.js';
 import log from './log.js';
 import ms from 'ms';
-import '@risingstack/trace';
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -188,12 +187,14 @@ function watch({ seq }) {
...
           // we do this by exiting and letting Heroku start over
           if (now - lastBootstrapped > c.timeToRedoBootstrap) {
-            stateManager.set({
-              seq: 0,
-              bootstrapDone: false,
-            }).then(() => {
-              process.exit(0);
-            });
+            stateManager
+              .set({
```
Haroenv: This is because i did that change in the last PR (#40) with the github web editor, and thus without commit hook.

PR #45, opened 2017-05-14
fix(memory): don't keep a reference of the `chain` in watch

Haroenv:
Reading up on promises and memory leaks, this might be the cause of our leak in the watch phase.

I'm not convinced that this is completely the same or that it will help, but we will see fast.


vvo:
Let's also use bluebird right now

Haroenv:
Will do after the weekend

PR #46, opened 2017-05-14
fix(memleak): in watch mode, do not use promise chain

vvo:
Promise chains are great for things that WILL resolve. If we know that
this will never resolve (like watch) then we should use a simple
callback based approach like queue/async.

This also re-ensures us that every changes are applied sequentially
(while with our previous check it was no more the case).

I hope this just works™


vvo:
image

:trollface:

PR #49, opened 2017-05-24
fix(changelog): fall back to master if the gitHead is undefined

Haroenv:
This happens from time to time for no particular reason, and in that case, we need to use 'master' as a fallback

### DIFF
--- a/github.js
+++ b/github.js
@@ -1,14 +1,7 @@
 import got from 'got';
 import race from 'promise-rat-race';
 
-function getChangelog({
-  githubRepo = {
-    user: '',
-    project: '',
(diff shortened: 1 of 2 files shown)


github.js
```diff
@@ -1,14 +1,7 @@
 import got from 'got';
 import race from 'promise-rat-race';
 
-function getChangelog({
-  githubRepo = {
-    user: '',
-    project: '',
-    path: '',
-  },
-  gitHead = 'master',
```
vvo: For reference, this is only called if gitHead key is not defined, not well null is passed.

PR #52, opened 2017-08-04
feat(quality): add a flag for very low quality packages

Haroenv:
This is used as a custom ranking, which acts much like the `deprecated` flag, in a way that it prefers packages with `false` as the value for `badPackage`. The organisations/owners that I chose to give a lower ranking are:

- [npmtest](https://yarnpkg.com/en/packages?q=npmtest-a&p=1)
- [npmdoc](https://yarnpkg.com/en/packages?q=npmdoc-a&p=1)

These authors are just duplicating packages in a way that provides no value alltogether.

### DIFF
--- a/config.js
+++ b/config.js
@@ -41,6 +41,7 @@ const defaultConfig = {
       'proximity',
       'attribute',
       'asc(deprecated)',
+      'asc(badPackage)',
       'desc(popular)',
       'exact',
       'custom',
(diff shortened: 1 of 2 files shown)


vvo:
Just checking, is there an npm tracking issue for this if it's spam? So that we remove this code at some point or add a comment in the code.

Also curious about which specific requests were impacted by those spammy packages?

PR #53, opened 2017-08-04
feat(schema): move git head into githubRepo

Haroenv:
This is cleaner as a structure. 

It's implemented backwards-compatible, and thus won't break the frontend, it should be fairly simple to migrate to this new structure once it's reindexed, and then we can remove the `gitHead` entry here.

In fact, I did this change on the website already, I only need to open the PR of [the branch](https://github.com/algolia/yarnpkg-website/tree/fix/move-head) once it's cleared here

### DIFF
--- a/github.js
+++ b/github.js
@@ -1,17 +1,17 @@
 import got from 'got';
 import race from 'promise-rat-race';
 
-function getChangelog({ githubRepo, gitHead }) {
+function getChangelog(githubRepo) {
   if (githubRepo === null) {
     return { changelogFilename: null };
   }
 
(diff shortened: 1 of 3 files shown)


rayrutjes:
I'm happy it is backward compatible ❤️

github.js
```diff
@@ -1,17 +1,17 @@
 import got from 'got';
 import race from 'promise-rat-race';
 
-function getChangelog({ githubRepo, gitHead }) {
```
vvo: You forgot to change the getChangeLog call, updating

PR #66, opened 2017-10-05
feat(typo): promote popular results to avoid typo squatting

Haroenv:
boostrap -> bootstrap

cc @julienbourdeau

### DIFF
--- a/config.js
+++ b/config.js
@@ -36,6 +36,7 @@ const defaultConfig = {
     disableTypoToleranceOnAttributes: ['keywords'],
     exactOnSingleWordQuery: 'attribute',
     ranking: [
+      'desc(popular)',
       'filters',
       'typo',
       'words',
@@ -43,7 +44,6 @@ const defaultConfig = {
       'attribute',
       'asc(deprecated)',
       'asc(badPackage)',
-      'desc(popular)',
       'exact',
       'custom',
     ],


Haroenv:
It works really well, I tried it in testing and in prod (will be overridden soon). 

query I found a bit tough is https://yarn.pm/ract

vvo:
Any concrete examples on the fact that it works really well? Would help understanding this.

Doesn't it have also drawbacks like searching for a particular package name won't show up because we first show the popular ones?

What if I am really searching for ract, will I find it?

Haroenv:
It works well for for example boostrap, which still shows up a few results lower, but bootstrap goes first. Typos very close to react are a bit of an issue because that exact result will only show up at page 10 or so because there are so many popular react packages. Ract was the only example I found here which was hard to find. Others with this issue don’t really exist. You can try this PR out by changing it in production, and it will revert to the real version in a few hours

vvo:
What I would do if feasible if query is `boostrap` is something like:

```
Results
---
2/ Exact match: Boostrap
1/ Most popular match with typo: Bootstrap
3+/ Other results
```

Knowing that in most cases, 1 and 2 would be the same record.
The tricky part of package search is knowing exactly how results are ranked. If I can ALWAYS rely on the fact that the top result is either a typo (google like "did you mean?") OR the exact match, that's better.

Otherwise if my exact match is always either 3rd, 4th, 5th without easy to understand rule that's hard. That's how npm search became tricky to understand to me. By trying to be smart, on 400000+ packages, it ended up being complex to understand the results and rely on them (= sometime I would find ract, sometime not = frustration and fear of using this search or relying on it).

julienbourdeau:
`react` makes it really hard :/ It seems like in the top 1% we want to boost, half of it will be related to react. I'm not sure what to do.

Haroenv:
I wonder how we could solve this

Haroenv:
This has negative aspects which make it too hard to find packages which are very close to popular one. Closing for that reason

PR #99, opened 2018-01-15
refactor: move into src folder and add basic tests

Haroenv:
Coverage isn't high yet, but at least it's testing _some_ of the things now, without doing any integration / network tests for now yet.

Also put all source files in `/src` because it is simpler like that


__mocks__/dotenv.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable import/no-commonjs */
```
vvo: You can define the env to be nodejs in eslint (or create an algolia/node subconfig too)

package.json
```diff
@@ -7,11 +7,11 @@
     "url": "https://www.algolia.com"
   },
   "scripts": {
-    "dev": "babel-node index.js",
+    "dev": "babel-node src/index.js",
```
vvo: Side question, do we still need babel-node you think? We might just want to avoid using import/export and go with require.

So we can drop babel* for this nodejs project. Using latest nodejs native will allow for better stack traces, which is way more important than being able to "use" (fake to) import. I know I did that but I want your opinion on it.

src/__tests__/config.test.js
```diff
@@ -0,0 +1,9 @@
+import config from '../config';
+
+it('gets the correct keys from env variables', () => {
```
vvo: 👍

src/__tests__/formatPkg.test.js
```diff
@@ -0,0 +1,32 @@
+import formatPkg from '../formatPkg';
+import rawPackages from './rawPackages.json';
+import isISO8601 from 'validator/lib/isISO8601';
+
+it('transforms correctly', () => {
```
vvo: Awesome to finally freeze a bit this code to allow easy diff when upgrading

package.json
Haroenv: We can use normal node now that it has rest/spread for objects!

screen shot 2018-01-16 at 09 54 00

__mocks__/dotenv.js
Haroenv: putting the env doesn't disable `import/no-commonjs`

package.json
Haroenv: @vvo I won't use require, since it's not worth it to change it now that import will be supported so soon in node. Babel is fine for now

PR #101, opened 2018-01-16
Add repository info

leonardosnt:
Now it supports Github, Gitlab and Bitbucket

See https://github.com/yarnpkg/website/issues/754


src/formatPkg.js
```diff
@@ -222,6 +239,65 @@ function getHomePage(homepage, repository) {
...
+  const result = repository.match(
+    /^https?:\/\/(?:www\.)?((?:github|gitlab|bitbucket)).((?:com|org))\/([^/]+)\/([^/]+)(\/.+)?$/
+  );
+
+  if (!result || result.length < 4) {
+    return null;
+  }
+
+  return {
+    host: `${result[1]}.${result[2]}`,
```
Haroenv: Here (and above as well) it would be prettier if we destructured `result`

leonardosnt: You mean like this:

```js
  const [, domain, domainTld, user, project, path = ''] = result;
  
  return {
    host: `${domain}.${domainTld}`,
    user,
    project,
    path,
  };
```

Haroenv: yep, that seems nicer to me

leonardosnt: > and above as well

Where? do you mean in another function? which one?

Haroenv: I meant in `getGithubInfo`, to which you referred in the comment

Haroenv:
I'll give this a from 0 deployment, and it'll be completely reindexed in 8h or so

leonardosnt:
Great! 🎉

PR #104, opened 2018-01-22
Add support for Gitlab and Bitbucket changelogs

leonardosnt:


Haroenv:
Looks lovely! Can you add a test for one that uses history.md so we’re sure it works? I think browserify has that

leonardosnt:
Browserify uses `changelog.markdown`. I used `express/body-parser`. https://raw.githubusercontent.com/expressjs/body-parser/master/HISTORY.md

Haroenv:
Interesting use of Rewire, first time I see it being used. Why did you use that over jest mocks?

leonardosnt:
To be honest, I didn't know about jest mocks.
(I'm not sure if we can get non-exported functions/variables with it)

I'll take a look in the jest-mock documentation. Maybe I can refactor using jest mocks later.

Haroenv:
Yes, you can override globals with it, syntax is slightly more obvious IMO, in this case it would be 

```js
jest.mock(
  "got",
  url =>
    gotSnapshotUrls.has(url) ? Promise.resolve({ url }) : Promise.reject() // eslint-disable-line prefer-promise-reject-errors
);
```

PR #105, opened 2018-01-24
Make getOwner work with Gitlab and Bitbucket

leonardosnt:


src/formatPkg.js
```diff
@@ -145,13 +146,32 @@ function getLicense(cleaned) {
...
+        name: user,
+        avatar: `https://github.com/${user}.png`,
+        link: `https://github.com/${user}`,
+      };
+    }
+
+    if (repository.host === 'gitlab.com') {
+      return {
+        name: user,
+        link: `https://gitlab.com/${user}`,
```
Haroenv: wondering if we should use the npm avatar as a fallback (even though it will usually be wrong)

src/__tests__/__snapshots__/formatPkg.test.js.snap
```diff
@@ -137,10 +137,9 @@ Array [
     "objectID": "@atlaskit/input",
     "originalAuthor": undefined,
     "owner": Object {
-      "avatar": "https://gravatar.com/avatar/f5f2da89496e4ad75ac1f9d7166e6b99",
-      "email": "design-platform@atlassian.com",
-      "link": "https://www.npmjs.com/~atlaskit",
-      "name": "atlaskit",
+      "avatar": "https://bitbucket.org/account/atlassian/avatar",
+      "link": "https://bitbucket.org/atlassian",
```
Haroenv: email is gone here, but I think it's fine

Haroenv:
I'm getting the error in CI in production too, I think parts of npm are a little down

src/__tests__/__snapshots__/formatPkg.test.js.snap
leonardosnt: Yeah, that's because the email was from npm account, you kniow..

src/formatPkg.js
leonardosnt: Isn't a bad idea.

Haroenv: let's add it so we don't need to add conditionals in the frontend for it

PR #106, opened 2018-02-01
feat(babel): add a forced keyword to babel plugins

Haroenv:
This is useful for the babel repl which will be visible [here](https://deploy-preview-1518--babel.netlify.com/repl) + https://github.com/babel/website/pull/1518 (Yarn search change isn't visible yet, but live coded it with @xtuc)

Not all packages have a keyword, but to be installable they start with `babel-plugin-` or `@babel/plugin`


vvo:
Good one and good test, thanks!

Haroenv:
We can do something similar for Gatsby and Webpack probably

PR #125, opened 2018-02-12
fix(deletes): handle npm deletions

### LINKED ISSUE #57: Handle deleted packages
I never implemented the delete change event. I think there's a _deleted flag somewhere in the feed but it's not very well documented and I am not sure npm follows this.

vvo:
We had to change the way replicate is done, it was actually completely
broken since we "fixed" the memory leak.

The reasoning is to always use db.changes as an event emitter to allow:
- getting docs from couchdb
- avoiding memory leaks (when complete event is triggered, docs are
already gone from memory)

Co-authored-by: Haroen Viaene <haroen.viaene@algolia.com>

fixes #57


src/createStateManager.js
```diff
@@ -1,15 +1,14 @@
-import algoliaIndex from './algoliaIndex.js';
 import c from './config.js';
 
 const defaultState = {
   seq: c.seq,
   bootstrapDone: false,
-  bootstrapLastId: undefined,
+  bootstrapLastId: 'zzz',
```
Haroenv: should be `undefined` again

PR #126, opened 2018-02-12
fix(bootstrap): move to production only in bootstrap

vvo:


vvo:
wait that's not complete I believe

vvo:
Ok it's good I only thought we needed to send back the state in the resolve of bootstrap but its not needed. merging!

PR #127, opened 2018-02-13
refactor(npm): rewrite getDownloads using async/await

Haroenv:
This is one of the (fewer) bits which are tested, so if they still pass, it should work


Haroenv:
Found the code a bit odd to read when I tried to fake the downloads, wrote this with async/await but removed the part which always returned 0

vvo:
> but removed the part which always returned 0

Where's this change?

PR #129, opened 2018-02-13
 refactor(main): write the main control flow in an async function

Haroenv:
Not necessary, but found this easier to read

depends on #128

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -24,15 +24,21 @@ const { index: mainIndex, client } = createAlgoliaIndex(c.indexName);
 const { index: bootstrapIndex } = createAlgoliaIndex(c.bootstrapIndexName);
 const stateManager = createStateManager(mainIndex);
 
-setSettings(mainIndex)
-  .then(() => setSettings(bootstrapIndex))
-  .then(() => stateManager.check())
-  .then(bootstrap)
-  .then(() => stateManager.get())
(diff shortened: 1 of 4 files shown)


vvo:
conflict now

Haroenv:
conflict fixed.

src/index.js
```diff
@@ -24,15 +24,21 @@ const { index: mainIndex, client } = createAlgoliaIndex(c.indexName);
...
-  .then(() => stateManager.check())
-  .then(bootstrap)
-  .then(() => stateManager.get())
-  .then(replicate)
-  .then(() => stateManager.get())
-  .then(watch)
-  .catch(error);
+async function main() {
+  await setSettings(mainIndex);
+  await setSettings(bootstrapIndex);
```
vvo: Why setSettings on the bootstrap index?

Haroenv:
Makes sense! I'll change it to what you suggested

Haroenv:
I set the settings on the bootstrap index as well so that if you use the search in the bootstrap phase, it will be set up correctly, and as a real follower (for quick QA)

vvo:
Unsure about what's needed here?

Haroenv:
I never investigated what the memory impact is (see your comment about scope). Probably still gonna quickly refactor to be the new logic and merge next week

PR #130, opened 2018-02-13
fix(watch): avoid memleak by not piling up docs

vvo:
See https://pouchdb.com/api.html#changes return_docs option.

Never use return_docs: true in live mode, what we want is include_docs


Haroenv:
you mention `include_docs`, but not in the code, what’s it for?

vvo:
include_docs is already in the code

Haroenv:
just read the docs, makes sense!

PR #160, opened 2018-03-14
feat(deprecated): add the attribute for faceting

### LINKED ISSUE #159: How to filter out deprecated packages?
I already have one `<RefinementList>` for specifying Gatsby's keywords. I tried adding another one to filter deprecated packages `<RefinementList attributeName="deprecated" defaultRefinement={[false]} />`

This caused this error: `Warning: Failed prop type: Invalid prop `defaultRefinement[0]` supplied to "AlgoliaRefinementList(Translatable(RefinementList))"`

I tried putting `false` in quotes but then no packages were returned.

I'm using react-instantsearch@4

Haroenv:
This allows it to be filtered on!

fixes #159

### DIFF
--- a/src/__tests__/__snapshots__/config.test.js.snap
+++ b/src/__tests__/__snapshots__/config.test.js.snap
@@ -29,6 +29,7 @@ Object {
       "filterOnly(concatenatedName)",
       "searchable(keywords)",
       "searchable(owner.name)",
+      "deprecated",
     ],
     "customRanking": Array [
       "desc(downloadsLast30Days)",
--- a/src/config.js
+++ b/src/config.js
@@ -28,6 +28,7 @@ const defaultConfig = {
       'filterOnly(concatenatedName)' /* optionalFacetFilters to boost the name */,
       'searchable(keywords)',
       'searchable(owner.name)',
+      'deprecated',
     ],
     customRanking: ['desc(downloadsLast30Days)'],
     disablePrefixOnAttributes: ['keywords', 'owner.name', 'owners.name'],


KyleAMathews:
Thanks!

But... this doesn't seem to be working yet — I merged in this PR https://github.com/gatsbyjs/gatsby/pull/4554 and if you search for "typegen" on gatsbyjs.org's package search (https://www.gatsbyjs.org/packages/?=typegen), the packages still show up even though I deprecated them all. Looking at the raw data coming from algolia in the network tab, it says `deprecated: false` for these packages still. Is it possible that the deprecated state isn't being pulled over? Or is there some way of verifying that the packages are actually deprecated on NPM?

vvo:
@KyleAMathews which package is that exactly? I tried to look for http://registry.npmjs.com/gatsby-typegen-filesystem and it does not says "deprecated" in here.

Let me know how it goes.

KyleAMathews:
Oh weird... so I guess NPM doesn't support deprecating all versions anymore. I ran `npm deprecate gatsby-typegen-filesystem "Replaced by gatsby-source-filesystem"` and nothing happened. But using your link @vvo, I trying specifying the version tagged with "latest" and now that version says it's deprecated and npm-search is now correctly not returning it. I'll check if I can just deprecate something with package-name@latest or if not, I'll look up the last version of packages and explicitly deprecate it. If there is a way to deprecate everything still, I'd love a pointer.

vvo:
> I'll look up the last version of packages and explicitly deprecate it

You have to deprecate package version by package version I believe. I am unsure so maybe ask npm support (they do answer in a very fast way)

PR #168, opened 2018-03-28
chore(npm): cleaner error logs

Haroenv:
right now we say "something went wrong asking for the downloads of", but there's actually downloads, scoped downloads and dependents,  which are three separate things

### DIFF
--- a/src/npm.js
+++ b/src/npm.js
@@ -14,13 +14,12 @@ export function info() {
   }));
 }
 
-const suppressError = ({ error, things, packages, returnValue }) => {
+const logError = ({ error, type, packages }) => {
   log.warn(
-    `Something went wrong asking the ${things} for \n${packages.join(
+    `Something went wrong asking the ${type} for \n${packages.join(
       ','
     )} \n${error}`
   );
-  return returnValue;
 };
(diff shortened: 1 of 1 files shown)


src/npm.js
```diff
@@ -56,14 +56,28 @@ export async function getDownloads(pkgs) {
     ...pkgsNamesChunks.map(pkgsNames =>
       got(`${c.npmDownloadsEndpoint}/point/last-month/${pkgsNames}`, {
         json: true,
-      }).catch(e => suppressError(e, pkgsNames, { body: {} }))
+      }).catch(error =>
+        suppressError({
```
vvo: Can we use: suppressError => logWarning and things => type?

Haroenv: Good idea, will refactor :)

vvo:
Would call it logWarning since we do not consider this an error, just a warning. I'd consider anything an error something that we cannot recover from (nitpicking here but will maybe more clearly give the intent)

vvo:
Tests are failing for now

Haroenv:
build passes again, was failing because npm was down 🙃

PR #169, opened 2018-03-29
feat(third-party): add handling of Angular CLI schematics, and rework registry subset

### LINKED ISSUE #167: Add filterable library-specific subsets in a future-proof way
```js
const packageDetail = {
  ...otherKeys,
  subset: {
    babel: "6",
    "angular-cli": "/path-to/scaffold"
    "gatsby": "component"
  }
}
```

Now I should still check how we can easily make it facetable, I'm not sure if putting it like this will allow that. An idea I had to solve that was this: 

```js
const packageDetail = {
  ...otherKeys,
  subset: {
    "keys": ["babel", "angular-cli", "gatsby"],
    babel: "6",
    "angular-cli": "/path-to/scaffold"
    "gatsby": "component"
  }
}
```

In this case only the `subset.keys` would be facetable, but it would mean that we can't use the name `keys` for any of the subsets (seems fine for me).

Haroenv:
This way we don't change the keywords anymore, and can do actual tests on the package contents.

cc @manekinekko :raised_hands:

@vvo I forgot which name you suggested instead of registry subset

closes #167


src/formatPkg.js
```diff
@@ -53,7 +53,9 @@ export default function formatPkg(pkg) {
 
   const owner = getOwner(repository, lastPublisher, author); // always favor the repository owner
   const badPackage = isBadPackage(owner);
-  const keywords = getKeywords(cleaned);
+  const registrySubsets = getRegistrySubsets(cleaned, pkg);
+  const __keywords = getKeywords(cleaned);
+  const keywords = [...__keywords, ...registrySubsets]; // concat with the subset for backward compat
```
vvo: Why not just `...getKeywords(cleaned)`?

src/formatPkg.js
```diff
@@ -86,6 +88,7 @@ export default function formatPkg(pkg) {
     homepage: getHomePage(cleaned.homepage, cleaned.repository),
     license,
     keywords,
+    registrySubsets,
```
vvo: If we only have an array of keywords for now and not actual metadata then let's just call it "computedKeywords".

Haroenv: I'm not sure about the name `keywords`

Haroenv: technically they are more like identifiers than keywords

Haroenv:
How can we have the last one, but not accidentally every package that has something to do with an unrelated kind of schematics? I only looked at the package.json key right now.

vvo:
Can you summarise what this PR changes in term of the response schema? Right now if I get it well it does adds a new property containing computed keywords. But seems like we want to go further and be able to add:

```js
{
  "name": "algoliasearch",
  "computedKeywords": ["angular", "schematics"], // computed by npm-search, not original data, used for faceting/filtering
  "computedMetadata": { // used for displaying
    "schematics"
  }
}
```

Is that correct and what we want? ^

vvo:
@manekinekko what's the search UI you are going to provide? It can help seeing it as for usecases and filtering.

If you filter on angular-schematics to only get this subset, all your usecase should work. @Haroenv Not sure to get why last one would not work.

Haroenv:
what's done in this PR will make a subset from the index to only have the packages which have the key `schematics`, but your example `c` wouldn't have that. Which kind of pattern do you think would match that too?

vvo:
Then "having the shematics metadata" must be translated into an actionnable filter, like:

```json
{
  "name": "algoliasearch",
  "computedKeywords": ["angular", "schematics"], // computed by npm-search, not original data, used for faceting/filtering
  "computedMetadata": { // used for displaying
    "hasSchematics": true,
    "schematics": "/some/path.json"
  }
}
```

Because there's no filter in Algolia that checks for attribute presence or not.

manekinekko:
I'm not familiar with actionable filters, but that sounds great ;)

Haroenv:
All you'll need to do instead of `index.search({ query, facetFilters: ['computedKeywords:angular-cli-schematic']})` (it's a Promise, so no need for a callback unless you really want) @manekinekko

Haroenv:
The actual schematics will be in `computedMetadata.schematic` by the way (but it's not completely reindexed yet)

PR #178, opened 2018-04-09
feat(ranking): do tie breaking based on the magnitude of downloads

### LINKED ISSUE #68: Custom ranking must be based on bucketed downloads

Haroenv:
Doing tie-breaking on the magnitue of the downloads (length of the number) allows for us to put other rules that simply "oh this one is 1 time more downloaded so it should be first"

This also introduced the key `_searchInternal` for data which isn't supposed to be rendered ever. I could put them in `unretrievableAttributes`, but that might be overkill.

We can test this by doing seq=0 and seeing what is happening in the bootstrap index, except if someone else has a better idea of testing it

fixes #68


vvo:
> I could put them in unretrievableAttributes, but that might be overkill.

Actually this might be a good idea for the index to not grow forever, thoughts?

src/npm.js
```diff
@@ -96,16 +96,20 @@ export async function getDownloads(pkgs) {
       : 0;
     const downloadsRatio = downloadsLast30Days / totalNpmDownloads * 100;
     const popular = downloadsRatio > c.popularDownloadsRatio;
-    // if the package is popular, we copy its name to a dedicated attribute
-    // which will make popular records' `name` matches to be ranked higher than other matches
-    // see the `searchableAttributes` index setting
-    const popularAttributes = popular ? { popularName: name } : {};
+    const downloadsMagnitude = downloadsLast30Days.toString().length;
```
vvo: Very good simple trick to use the string length :)

PR #179, opened 2018-04-11
feat(settings): allow to make a PR which changes both the settings and the data

Haroenv:
We need this change, because right now, if we merge a PR which changes the index settings, they are immediately applied to the production index, which could not have the correct data yet

Flow before this PR:

1. setSettings on production index
2. setSettings on bootstrap index
3. bootstrap
4. copy production settings to bootstrap index
5. move bootstrap index to production index
6. delete bootstrap index
7. replicate
8. watch

Flow after this PR

1. setSettings on bootstrap index
3. bootstrap (will set settings to prod if up to date)
5. move bootstrap index to production index with its settings
6. delete bootstrap index
7. replicate
8. watch


src/index.js
```diff
@@ -148,14 +149,13 @@ async function bootstrap(state) {
 
 async function moveToProduction() {
   log.info('🚚 starting move to production');
-  await client.copyIndex(c.indexName, c.bootstrapIndexName, [
```
Haroenv: we no longer copy the settings from the prod index to the bootstrap index, but rely on the bootstrap index being correct

src/index.js
```diff
@@ -83,6 +83,7 @@ function infoDocs(offset, nbDocs, emoji) {
 
 async function bootstrap(state) {
   if (state.seq > 0 && state.bootstrapDone === true) {
+    await setSettings(mainIndex);
```
Haroenv: setSettings will always be called when bootstrap is done

src/index.js
```diff
@@ -28,7 +28,7 @@ const { index: mainIndex, client } = createAlgoliaIndex(c.indexName);
 const { index: bootstrapIndex } = createAlgoliaIndex(c.bootstrapIndexName);
 const stateManager = createStateManager(mainIndex);
 
-setSettings(mainIndex)
+Promise.resolve()
```
vvo: If only someone did a chore PR of moving to async/await... :D

src/index.js
```diff
@@ -148,14 +149,13 @@ async function bootstrap(state) {
 
 async function moveToProduction() {
   log.info('🚚 starting move to production');
-  await client.copyIndex(c.indexName, c.bootstrapIndexName, [
+
+  const currentState = await stateManager.get();
+  await client.copyIndex(c.bootstrapIndexName, c.indexName, [
```
vvo: I believe we should write `await client.copyIndex(c.bootstrapIndexName, c.indexName)` which will copy the data + all settings (+synonyms, query rules).

The current line will not copy the data

Haroenv: oh you're right, good of you for catching that!

PR #181, opened 2018-04-13
feat(yeoman): Identify yeoman generators through computedKeywords

pixelastic:
This follows the official rules of what is considered a yeoman
package:
- name must start with `generator-`
- `yeoman-generator` must be in the keywords

When that happens, it will add `generator` to the `computedKeywords`.

### DIFF
--- a/src/formatPkg.js
+++ b/src/formatPkg.js
@@ -215,7 +215,7 @@ const registrySubsetRules = [
   }),
 
   ({ name, keywords = [] }) => ({
-    name: 'generator',
+    name: 'yeoman-generator',
     include:
       name.startsWith('generator-') && keywords.includes('yeoman-generator'),
   }),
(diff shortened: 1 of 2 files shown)


pixelastic:
Does it need a new full re-indexing to take effect once merged?

Haroenv:
Yes, this will require a full reindexing (of which one is being done _right now_)

src/formatPkg.js
```diff
@@ -214,6 +214,12 @@ const registrySubsetRules = [
     include: /^(@vue\/|vue-|@[\w-]+\/vue-)cli-plugin-/.test(name),
   }),
 
+  ({ name, keywords = [] }) => ({
+    name: 'generator',
```
Haroenv: I'd call this `yeoman-generator`

pixelastic: Good call

PR #188, opened 2018-04-26
feat(relevance): put name, description and eywords on same level

Haroenv:
This will give significantly more relevant results (you can try it out with the index `npm-search-relevance-test` I created to test out these settings.

I had to remove the settings to disable exact and prefix from those attributes, because it otherwise clashes with errors like

```
invalid setting for disablePrefixOnAttributes, you should requests all attributes with same priority: attr1, attr2,  does not exists
```

cc @redox, since you did the original ranking decisions here :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,11 +147,11 @@ We're restricting the search to use a subset of the attributes only:
 
 ##### Prefix Search
 
-Algolia provides default prefix search capabilities (matching words with only the beginning). This is disabled for the `keywords`, `owner.name` and `owners.name` attributes.
+Algolia provides default prefix search capabilities (matching words with only the beginning). This is disabled for the `owner.name` and `owners.name` attributes.
 
 ##### Typo-tolerance
 
-Algolia provides default typo-tolerance. Typo-tolerance is disabled for the `keywords` attribute.
+Algolia provides default typo-tolerance.
 
 ##### Exact Boosting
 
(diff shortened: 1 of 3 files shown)


Haroenv:
cc @manekinekko too, you can try with this index for now, I think the relevance problem you mentioned is fixed :)

Haroenv:
fixes `{filters: 'keywords:angular', query: 'i18n'}` and `{ query: 'postgresql' }`

redox:
What were the broken/faulty queries? Some of those changes appear strange to me but maybe the `_searchInternal.popularName` attribute is enough to address all of those.

manekinekko:
this sample query: https://www.npmjs.com/search?q=angular%20i18n should show these TOP two packages:

- `@ngx-translate/core`
- `@ngx-translate/http-loader`

These two packages match the best the query. Please note that these 2 packages don't have `i18n` in their names. 

The same query using https://ngx.tools/#/search?q=i18n (which is using this search index) shows different results.

Haroenv:
`@ngx-translate/http-loader` doesn't have `i18n`, anywhere in the keywords or description, I guess some synonyms could help though

manekinekko:
I think we can leave out packages that don't meet the search queries (ie. `i18n`). I mean, at some point, module authors would have to include and maintain the right keywords that describe the best the use of their package.

So, if the `@ngx-translate/http-loader` doesn't include the search query in any of its properties, it should be downranked. WDYT? 

Also, could you elaborate on the synonyms topic? I'm curious to know how they are handled in the index?

Haroenv:
See the next unmerged PR, there’s currently only a synonym for _ to underscore

PR #191, opened 2018-04-30
feat(relevance): add another query rule on name

Haroenv:
We earlier added a query rule on concatenated name to promote if that matched, but for a query like `react-native`, you still want that first.


Haroenv:
This helps for the query `react-native` (note that it's not `react native`, maybe we need a third query rule that does split name where we replace special characters with spaces)

PR #193, opened 2018-04-30
feat(relevance): add query rule on alternative names

Haroenv:
This is useful for pretty much all names with dashes, if you'd search them with spaces in place of dashes, where you'd still want them to match exactly. This is visible if there's another package which
also has that in the name, but is more popular
Example queries where the exact match now ranks higher are:

- react native --> `react-native` should be first, but is fourth
- (couldn't find more examples, but I'm sure some can be found, please comment with it)

Related to #191, but mergeable individually


Haroenv:
another example: 

* absolute path --> `absolute-path` should be first, but is third

PR #195, opened 2018-05-04
feat(relevance): enable alternative names query rule

Haroenv:
see #194 for info and queries to try

Waiting for a complete reindex (next Friday) to try this out


Haroenv:
tested it and works great!

PR #221, opened 2018-06-12
feat(query rules): add filtering on attr:value

Haroenv:
Added it on `author:{owner.name}`, `owner:{owner.name}` and `keyword:{keywords}`. Do you have more suggestions @vvo?


vvo:
Do you have example queries where this would lead to better results? Only trying to get it right :)

Haroenv:
`author:algolia instantsearch` would be the same as filtering manually by clicking

PR #242, opened 2018-07-12
Replication to watch stage bug fix

JureSotosek:
Fixes a bug which causes the "watch" stage to begin before all the documents in the "replication" cargo queue have been processed/saved. Because those two queues end up saving documents at the same time it could cause changes to be applied out of order and consequently cause wrong versions of npm documents to be index in the Algolia index.

To demonstrate the bug, here is a log from the old version:
```
Jul 12 09:07:14 npm-search[55176] INFO:  🗿 npm ↔️ Algolia replication starts ⛷ 🐌 🛰
Jul 12 09:07:24 npm-search[55176] INFO:  ⛷ Bootstrap: done
Jul 12 09:07:24 npm-search[55176] INFO:  🐌 Replicate: Asking for 10 changes since sequence 5499870
Jul 12 09:07:35 npm-search[55176] INFO:  🔍 Found and saved 7 packages
Jul 12 09:07:35 npm-search[55176] INFO:  🐌 Synced 5499877/5499880 changes (99%), current rate: 0 changes/s (1m remaining)
Jul 12 09:07:35 npm-search[55176] INFO:  🐌 We reached the npm current sequence
Jul 12 09:07:35 npm-search[55176] INFO:  🛰 Watch: 👍 We are in sync (or almost). Will now be 🔭 watching for registry updates, since 5499877
Jul 12 09:07:36 npm-search[55176] INFO:  🔍 Found and saved 3 packages
Jul 12 09:07:37 npm-search[55176] INFO:  🐌 Synced 5499880/5499880 changes (100%), current rate: 1 changes/s (0ms remaining)
Jul 12 09:07:37 npm-search[55176] INFO:  🐌 We reached the npm current sequence
Jul 12 09:07:37 npm-search[55176] INFO:  🔍 Found and saved 1 packages
Jul 12 09:07:37 npm-search[55176] INFO:  🛰 Synced 5499878/5499880 changes (99%), current rate: 2 changes/s (1s remaining)
Jul 12 09:07:38 npm-search[55176] INFO:  🔍 Found and saved 1 packages
Jul 12 09:07:39 npm-search[55176] INFO:  🛰 Synced 5499879/5499881 changes (99%), current rate: 1 changes/s (3s remaining)
```
You can see that when "🐌 We reached the npm current sequence" is logged, watch stage begins, but after that "🐌 We reached the npm current sequence" is logged again. That is because there were documents still left in the "replicate" queue and watch stage began, before queue has been "drained".

My solution was to call resolve() only after the queue has been drained and changes.on('change) has registered that the npmSeqToReach has been reached.

Note: I talked to @Haroenv about it.


src/index.js
```diff
@@ -188,16 +189,17 @@ async function replicate({ seq }) {
...
           })
         )
-        .then(({ seq: lastDocSeq }) => {
-          if (lastDocSeq >= npmSeqToReach) {
-            log.info('🐌 We reached the npm current sequence');
-            changes.cancel();
-          }
-        })
-        .then(done)
+        .then(() => done())
```
Haroenv: this could stay `.then(done)`

JureSotosek: No, because then err would be triggered, look at the call back supplied to the queue

Haroenv: Not sure if I get the difference, this both is the same:

```js
Promise.resolve('hi').then(console.log);
// identical
Promise.resolve('hi').then(msg => console.log(msg));
```

JureSotosek: The callback you supply to the queue is (`done`):
```
if (err) {
  reject(err)
}
```
So if there is any argument given to the function `done` it is going to throw an error.

```
.then(() =>
          stateManager.save({
            seq: docs[docs.length - 1].seq,
          })
        )
        .then(done)
        .catch(done);
```

If you did it this way, `stateManager.save` if going to resolve with `state`, and that `state` is going to get passed to the `done` function as an argument and `done`, because it just does `if (err)`, is going to throw an error.

Haroenv: ah okay, good call, thanks

JureSotosek:
Totally forgot about this 😅

Haroenv:
me too, but I tried it out, and seemed to be better indeed if those conditions are forced

Issue #187, opened 2018-04-24
[FEATURES] add additional information for a package


manekinekko:
Hi guys. I've got requests for some new additions to the `npm-search` index. Let's discuss them here:

- [ ] open issues on gihtub: How many open issues are there.
- [ ] number of gihtub stars: How many people stared this projet.
- [ ] number of gihtub forks: How many people stared this projet.
- [ ] open PR on gihtub: How many open PRs are there.
- [ ] last commit on gihtub: When was the last commit on master.
- [ ] add any other usefull information (here are all the properties returned by [Github's API](https://developer.github.com/v3/repos/#get))

Does this make sense?

Haroenv:
It would make sense to also query GitHub for those for which we have the data. In the Yarn website we do this frontend once the detail page has been requested. For requesting GitHub data we'll need to have some API key rotation like npms does: [source](https://github.com/npms-io/npms-analyzer/blob/238a7480e8beb6b57ba27b80593a67b3a3f75b9a/lib/analyze/collect/github.js#L6). 

It would also make the replication slower, but that would be fine IMO, it's just the API usage limitations that I wasn't able to overcome for now

MartinKolarik:
One thing that I would also love to add is jsDelivr hits in the last month. Right now the search results are sorted by npm downloads, which is great for nodejs/backend packages, but doesn't work that well for browser/frontend packages, especially those, which recommend CDN as the primary installation option. 

It would be nice if we could either combine those numbers somehow, or simply had an option for sorting by cdn hits rather than npm downloads. Implementation should be as easy as using either [this](https://data.jsdelivr.com/v1/package/npm/algoliasearch/stats/month) or [this](https://data.jsdelivr.com/v1/stats/packages/month) endpoint of our API.

Haroenv:
Is it possible to get the monthly downloads batched, like [this](https://data.jsdelivr.com/v1/stats/packages/month) but with a specified 100/200 packages to look this up from at the same time, so we can look them all up?

MartinKolarik:
There's ?page, so you can just do ?page=1, ?page=2, ... until you get no results.

Haroenv:
I won't have time for now to add this, but feel free to contribute or contact me if I can help. It indeed could be possible to augment the index like that, although the flow we currently have is 

1. get all packages in batches
2. loop over a batch of packages to get more info or get more info from the whole batch

So we'd need to control which packages would be in the batch (maybe something that can be added to your API first?

[...]

PR #263, opened 2018-08-22
feat(data): add jsDelivr hits

Drgy:
As discussed in #187

### DIFF
--- a/src/jsDelivr.js
+++ b/src/jsDelivr.js
@@ -1,0 +1,35 @@
+import got from 'got';
+import c from './config.js';
+import log from './log.js';
+
+const hits = new Map();
+
+function formatHits(pkg) {
+  if (pkg.type !== 'npm') {
+    return;
+  }
+
+  hits.set(pkg.name, pkg.hits);
+}
+
+export async function loadHits() {
+  const hitsJSONpromise = got(c.jsDelivrHitsEndpoint, { json: true }).catch(
+    error => {
+      log.warn(
+        `Can't download hits data from ${
+          c.jsDelivrHitsEndpoint
+        }, error: ${error}`
+      );
+    }
+  );
+
+  const hitsJSON = (await hitsJSONpromise).body;
+  hits.clear();
+  hitsJSON.forEach(formatHits);
+}
+
+export function getHits(pkgs) {
+  return pkgs.map(({ name }) => ({ jsDelivrHits: hits.get(name) || 0 }));
+}
+
+setInterval(loadHits, 24 * 60 * 60 * 1000);
--- a/src/jsDelivr.js
+++ b/src/jsDelivr.js
@@ -31,5 +31,3 @@ export async function loadHits() {
 export function getHits(pkgs) {
   return pkgs.map(({ name }) => ({ jsDelivrHits: hits.get(name) || 0 }));
 }
-
-setInterval(loadHits, 24 * 60 * 60 * 1000);
--- a/src/jsDelivr.js
+++ b/src/jsDelivr.js
@@ -1,6 +1,5 @@
 import got from 'got';
 import c from './config.js';
-import log from './log.js';
 
 const hits = new Map();
 
@@ -13,16 +12,7 @@ function formatHits(pkg) {
 }
 
 export async function loadHits() {
-  const hitsJSONpromise = got(c.jsDelivrHitsEndpoint, { json: true }).catch(
-    error => {
-      log.warn(
-        `Can't download hits data from ${
-          c.jsDelivrHitsEndpoint
-        }, error: ${error}`
-      );
-    }
-  );
-
+  const hitsJSONpromise = got(c.jsDelivrHitsEndpoint, { json: true });
   const hitsJSON = (await hitsJSONpromise).body;
   hits.clear();
   hitsJSON.forEach(formatHits);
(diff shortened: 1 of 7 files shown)


Haroenv:
This is cool, thanks!

MartinKolarik:
@Haroenv we didn't make any index config changes as I thought that's best left up to you but we'd want to:
 - use this for sorting results
 - ideally be able to say "npm downloads are more important" vs "jsDelivr hits are more important" on search time

Haroenv:
cc @martinkolarik, to see if this is proper usage of the API (seeing that we will be calling it ±1000000 times spread over a day and a few thousand times as well regularly

MartinKolarik:
@Haroenv yes, I asked @Drgy to do this and already reviewed. And you only call it once per 24 hours ;)

Haroenv:
I didn’t look at the implementation yet, note that the process regularly gets shut down and started again, so any timers won’t work :) I suggest it to follow the same flow as requesting things from npm

MartinKolarik:
I believe this will work fine but let me know what you think after you check the code. It loads all stats on startup (there's no perf difference in getting stats for all packages vs just a hundred) and then updates the data every 24 hours if the process still runs.

src/npm.js
```diff
@@ -14,11 +14,9 @@ export function info() {
   }));
 }
 
-const logWarning = ({ error, type, packages }) => {
```
MartinKolarik: @Haroenv just a note that changes in this file are a bugfix not related to jsDelivr. This function expected `packages` to be an array but it was actually a string everywhere so it would fail on `packages.join()`

src/jsDelivr.js
```diff
@@ -0,0 +1,35 @@
...
+  const hitsJSON = (await hitsJSONpromise).body;
+  hits.clear();
+  hitsJSON.forEach(formatHits);
+}
+
+export function getHits(pkgs) {
+  return pkgs.map(({ name }) => ({ jsDelivrHits: hits.get(name) || 0 }));
+}
+
+setInterval(loadHits, 24 * 60 * 60 * 1000);
```
Haroenv: nit: `ms('1 day')` is preferable here

src/jsDelivr.js
```diff
@@ -0,0 +1,35 @@
+import got from 'got';
+import c from './config.js';
+import log from './log.js';
+
+const hits = new Map();
```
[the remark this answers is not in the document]
MartinKolarik: It *is* called in init phase, the side-effect here is only the `setInterval` which runs later to refresh the data. I agree that it isn't so nice but wasn't sure if there's a better place to put it.

It's less than 1 MB in JSON, I'd expect not much more in memory.

src/jsDelivr.js
```diff
@@ -0,0 +1,35 @@
...
+  if (pkg.type !== 'npm') {
+    return;
+  }
+
+  hits.set(pkg.name, pkg.hits);
+}
+
+export async function loadHits() {
+  const hitsJSONpromise = got(c.jsDelivrHitsEndpoint, { json: true }).catch(
+    error => {
```
MartinKolarik: I was wondering if it would be better to reject with the error here because if this fails at the start there won't be hits for any packages.

src/jsDelivr.js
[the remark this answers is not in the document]
Haroenv: I thought we were talking about a significantly bigger amount of data, 1MB shouldn't be a problem

src/jsDelivr.js
Haroenv: yes, we should probably reject here, since otherwise every package will log if this can't be loaded

src/jsDelivr.js
Haroenv: also, we do a weekly "bootstrap", where we throw away all data and start from scratch. It's probably enough to have data up to date weekly, rather than adding a timer IMO

MartinKolarik:
@Haroenv anything else we should address?

Haroenv:
Nothing particular, I just haven't had the time to try this out (see the perf changes)

MartinKolarik:
Ok 😄

@Drgy one thing we should probably do is add this to the schema in the readme: https://github.com/algolia/npm-search#schema

Haroenv:
I'll try this first thing tomorrow

MartinKolarik:
Awesome.

Haroenv:
seems to be working ok so far, will see how it changes over the coming 20-ish hours the replication takes. Some packages already have the jsDelivr number there

Haroenv:
Fully live now

Haroenv:
All live now, you're good to use `jsDelivrHits` on your frontend now @MartinKolarik @Drgy

MartinKolarik:
@Haroenv thanks! Now the followup question is, how do we use this as an additional ranking criteria?

Haroenv:
We'd have to set up a replica with that different ranking. I'll take a look at that soon-ish, since I need to discuss if there's place on the machine we're using for a double in records :)

MartinKolarik:
Would it be possible to use jsDelivr hits and npm downloads at the same time? I'd like to achieve something like this for packages with the same textual relevance:

1. Packages with many jsDelivr hits and many npm downloads.
2. Packages with many jsDelivr hits and few npm downloads.
3. Packages with few jsDelivr hits and many npm downloads.
4. Packages with few jsDelivr hits and few npm downloads.

Alternatively, the 2nd and 3rd position could be swapped in which case this might make sense even for the main index and remove the need of a replica.

PR #296, opened 2018-12-07
feat(keywords): add webpack-scaffold

evenstensberg:
Adds webpack scaffold to lookup :)


Haroenv:
Thanks for adding this so swiftly!

Two questions: 

- Are you sure you want to enforce the keyword?
- Can you add a small test in the style of the other computed keywords?

evenstensberg:
Like dat 👍

evenstensberg:
Tests were failing locally fwiv, not related to my changes

Haroenv:
Yes, I fixed that on master, thanks for contributing!

This change will be live in about a week (unless you want it earlier, but it will still take at least 24hr)

Haroenv:
I’ll let you know when it’s done. In the mean time, any package that matches and is updated since now will already be filterable

evenstensberg:
Gotcha!

PR #303, opened 2019-01-03
feat(state): use Redis instead of Algolia

Haroenv:
Since apparently this is now a "bad practice", I use Heroku Redis for this. I tested it locally and the data gets persisted as was expected.

I'll talk to Clement to see if this is really necessary and if maybe the queue can be cleared


Haroenv:
So I have talked to Clement, the situation we are having will be resolved automatically when the bug in the engine has been released, which will be early next week. We can still consider to merge this PR if it proves not to work

vvo:
To ensure we make merge this, it needs:
- simple docker compose file like here: https://docs.docker.com/compose/compose-file/#compose-file-structure-and-examples
- update readme instructions on how to run locally

thoughts?

Haroenv:
It works with `heroku local` without having to lose my mind with docker etc. As said in the "what to do" document, I hope this solution won't be necessary in the first place 🎃

Haroenv:
The index is up to date and updating again as expected, so this change isn't needed

PR #320, opened 2019-02-19
Include directory details from repository objects

greysteil:
Now that https://github.com/npm/rfcs/pull/19 and https://github.com/npm/cli/pull/140 have been merged (woo! yay!) a number of packages have started specifying their directory within a monorepo as a `directory` attribute on their `repository` object. See, for example, [react-dom](http://registry.npmjs.org/react-dom/16.8.2).

This PR update the `getRepositoryInfo` logic to pull path details from this directory field, if present (and fall back to trying to pull it out of the URL). If merged, it will mean the GitHub link on pages like [this one](https://yarnpkg.com/en/package/react-dom) link to the correct folder within the react monorepo (woo! yay!).

(Credit to @ybiquitous for pointing me in the right direction on this one.)

### DIFF
--- a/src/__tests__/formatPkg.test.js
+++ b/src/__tests__/formatPkg.test.js
@@ -252,6 +252,18 @@ describe('test getRepositoryInfo', () => {
       url: 'git+https://bitbucket.org/2klicdev/2klic-sdk.git',
     };
 
+    const githubRepoWithDirectory = {
+      type: 'git',
+      url: 'https://github.com/facebook/react.git',
+      directory: './packages/react-dom',
+    };
+
+    const githubRepoWithPathUrlAndDirectory = {
+      type: 'git',
+      url: 'https://github.com/facebook/react/tree/master/packages/wrong',
+      directory: './packages/react-dom',
+    };
+
     expect(getRepositoryInfo(githubRepo)).toEqual({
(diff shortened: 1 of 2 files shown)


greysteil:
Just the Yarn website.

I run Dependabot, which needs directory details for packages that are part of monorepos in order to create better PRs (so, for example, the link on `react-dom` in [this PR](https://github.com/parity-js/qr-signer/pull/49) is better now that I got the react team specifying directories). That's why I created the npm RFC on this stuff - especially as it should help other automated tools, too.

As part of the follow-up to getting that RFC accepted I'm going around all of the big monorepos getting them on board with the new best practice, and the Gatsby folks correctly mentioned that their current hack (including directory details in the URL) worked for the Yarn website, and others, whereas the new official setup wouldn't. Updating things here fixes that. 🙌

Do you know if the npmjs.com website is also powered by Algolia under the hood? It's closed source, so the one place I can't fix things, but if this changes fixes the repository link at e.g., https://www.npmjs.com/package/react-dom then that would be even better.

Regardless, thanks for the review!

Haroenv:
Npm has their own system, this is powering the yarn website, Gatsby, babel plugins, codesandbox and a few others. 

Thanks a lot for your contribution here!

PR #346, opened 2019-06-14
Adds whether there are types for a package to an index

orta:
Based [on this](https://twitter.com/mxstbr/status/1139498444777824256) discussion. This PR adds a new field to the JSON index for npm search

```sh
{
  ...
  "types": { "ts": "included" }
}
```

or

```sh
{
  ...
  "types": { "ts": "@types/lodash" }
}
```

and if they don't support it:

```sh
{
  ...
 "types": {  "ts": null }
}
```

which might be absent in JSON? Or false. Actually not sure. But definitely falsy.

Running `yarn format` did a few more changes than I would have done solo, but I'll leave comments in here about what I've changed.

Note: I've not been able to successfully run this locally against a set of packages, so right now it's somewhat ready for testing but not guaranteed to be shippable without risk. Should work tho - hah. 

Adding comments now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,94 +39,94 @@ For every single NPM package, we create a record in the Algolia index. The resul
 
 ```json5
 {
-  "name": "babel-core",
-  "concatenatedName": "babelcore",
-  "downloadsLast30Days": 10978749,
-  "downloadsRatio": 0.08310651682685861,
-  "humanDownloadsLast30Days": "11m",
-  "jsDelivrHits": 11684192,
-  "popular": true,
-  "version": "6.26.0",
-  "versions": {
+  name: 'babel-core',
+  concatenatedName: 'babelcore',
+  downloadsLast30Days: 10978749,
+  downloadsRatio: 0.08310651682685861,
+  humanDownloadsLast30Days: '11m',
+  jsDelivrHits: 11684192,
+  popular: true,
+  version: '6.26.0',
+  versions: {
     // [...]
-    "7.0.0-beta.3": "2017-10-15T13:12:35.166Z"
+    '7.0.0-beta.3': '2017-10-15T13:12:35.166Z',
   },
-  "tags": {
-    "latest": "6.26.0",
-    "old": "5.8.38",
-    "next": "7.0.0-beta.3"
+  tags: {
+    latest: '6.26.0',
+    old: '5.8.38',
+    next: '7.0.0-beta.3',
   },
-  "description": "Babel compiler core.",
-  "dependencies": {
-    "babel-code-frame": "^6.26.0"
-    // [...]
-  },
-  "devDependencies": {
-    "babel-helper-fixtures": "^6.26.0"
+  description: 'Babel compiler core.',
+  dependencies: {
+    'babel-code-frame': '^6.26.0',
     // [...]
   },
-  "repository": {
-    "url": "https://github.com/babel/babel/tree/master/packages/babel-core",
-    "host": "github.com",
-    "user": "babel",
-    "project": "babel",
-    "path": "/tree/master/packages/babel-core",
-    "branch": "master"
+  devDependencies: {
+    'babel-helper-fixtures': '^6.26.0',
+    // [...]
   },
-  "readme":
-    "# babel-core\n\n> Babel compiler core.\n\n\n [... truncated at 200kb]",
-  "owner": {
+  repository: {
+    url: 'https://github.com/babel/babel/tree/master/packages/babel-core',
+    host: 'github.com',
+    user: 'babel',
+    project: 'babel',
+    path: '/tree/master/packages/babel-core',
+    branch: 'master',
+  },
+  readme: '# babel-core\n\n> Babel compiler core.\n\n\n [... truncated at 200kb]',
+  owner: {
     // either GitHub owner or npm owner
-    "name": "babel",
-    "avatar": "https://github.com/babel.png",
-    "link": "https://github.com/babel"
+    name: 'babel',
+    avatar: 'https://github.com/babel.png',
+    link: 'https://github.com/babel',
   },
-  "deprecated": false,
-  "badPackage": false,
-  "homepage": "https://babeljs.io/",
-  "license": "MIT",
-  "keywords": [
-    "6to5",
-    "babel",
-    "classes",
-    "const",
-    "es6",
-    "harmony",
-    "let",
-    "modules",
-    "transpile",
-    "transpiler",
-    "var",
-    "babel-core",
-    "compiler"
+  deprecated: false,
+  badPackage: false,
+  homepage: 'https://babeljs.io/',
+  license: 'MIT',
+  keywords: [
+    '6to5',
+    'babel',
+    'classes',
+    'const',
+    'es6',
+    'harmony',
+    'let',
+    'modules',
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -39,94 +39,94 @@ For every single NPM package, we create a record in the Algolia index. The resul
 
 ```json5
```
orta: prettier updated this, I did add `ts` to the example though

src/__tests__/__snapshots__/config.test.js.snap
```diff
@@ -179,9 +179,11 @@ Object {
   "maxObjSize": 450000,
   "npmDownloadsEndpoint": "https://api.npmjs.org/downloads",
   "npmRegistryEndpoint": "https://replicate.npmjs.com/registry",
+  "npmRootEndpoint": "https://api.npmjs.org/",
   "popularDownloadsRatio": 0.005,
   "replicateConcurrency": 10,
   "seq": null,
   "timeToRedoBootstrap": 604800000,
+  "unpkgRoot": "https://unpkg.com/",
 }
```
orta: To make HEAD requests to know if something existed I needed access to these two API roots

src/formatPkg.js
```diff
@@ -61,7 +61,6 @@ export default function formatPkg(pkg) {
   const splitName = cleaned.name.replace(/[-/@_.]+/g, ' ');
 
   const tags = pkg['dist-tags'];
-
   const rawPkg = {
```
orta: Hah, err, looks like I or prettier were being pedantic.

src/npm.js
```diff
@@ -20,6 +20,13 @@ const logWarning = ({ error, type, packagesStr }) => {
   );
 };
 
+export function validatePackageExists(pkgName) {
+  return got(`${c.npmRootEndpoint}/${pkgName}`, {
+    json: true,
+    method: 'HEAD',
+  }).then(response => response.statusCode === 200);
+}
```
orta: we only need to know if it exists - so you can use `curl -I https://registry.npmjs.org/@types/lodash` to see if it exists

src/saveDocs.js
```diff
@@ -26,13 +27,15 @@ function addMetaData(pkgs) {
     getDependents(pkgs),
     getChangelogs(pkgs),
     getHits(pkgs),
-  ]).then(([downloads, dependents, changelogs, hits]) =>
+    getTSSupport(pkgs),
+  ]).then(([downloads, dependents, changelogs, hits, ts]) =>
```
orta: `ts[index]` here will always be ` { ts: %something% }`

src/typescriptSupport.js
```diff
@@ -0,0 +1,41 @@
...
+
+import { validatePackageExists } from './npm.js';
+import { fileExistsInUnpkg } from './unpkg.js';
+
+/**
+ * Basically either
+ *  - { ts: undefined }  for no existing TypeScript support
+ *  - { ts: "@types/module" } - for definitely typed support
+ *  - { ts: "included" } - for types shipped with the module
+ * */
```
orta: my pseudo typescript interface 🍡

Haroenv:
@mjackson, are you fine with us calling unpkg here this many times?

src/typescriptSupport.js
```diff
@@ -0,0 +1,41 @@
...
+  // The 2nd most likely is definitely typed
+  const defTypeName = `@types/${pkg.name}`;
+  const defTyped = await validatePackageExists(defTypeName);
+  if (defTyped) {
+    return { ts: defTypeName };
+  }
+
+  // Check if main's JS file can be resolved to a d.ts file instead
+  const main = pkg.main || 'index.js';
+  if (main.endWith('.js')) {
```
Haroenv: Is this not “endsWith”?

Haroenv:
Would be cool if there were tests for the new code :) looking at it, i think we’ll be fine with the impact on indexing time

src/typescriptSupport.js
orta: Oops, yep, fixed

orta:
It would do a head request to unpkg _at most_ once per package given that it's likely the majority of packages don't have types, then I'd expect the majority of packages to do this

Haroenv:
I think we should nest this inside a “types” property, so that if there’s interest, we can add flow & reason eg

orta:
Makes sense - it does that now, and has tests 👍

robb:
`null` might be a more convenient alternative to `undefined`, given that the latter isn't available in JSON…

orta:
Yeah, great point - fixed!

arcanis:
The only caveat is that the index for `foo` won't be recomputed when a new `@types/foo` package appears, am I correct? It will require to publish a new version of `foo` 🤔

Haroenv:
There’s a weekly refresh of the whole index to catch things like this. I guess we could add some extra code to schedule an update of the relevant package of an @types but it’s not that big of an issue I think

mjackson:
@Haroenv I'm 👍 with calling unpkg like this. I'll let y'all know as soon as I launch a better way to do this than making `HEAD` requests...

Haroenv:
Perfect 👌 I’ll take a look at if this works well next week

README.md
```diff
@@ -39,94 +39,94 @@ For every single NPM package, we create a record in the Algolia index. The resul
...
-  "changelogFilename": null, // if babel-core had a changelog, it would be the raw GitHub url here
-  "objectID": "babel-core",
-  "_searchInternal": {
-    "popularName": "babel-core",
-    "downloadsMagnitude": 8,
-    "jsDelivrPopularity": 5
-  }
+  lastCrawl: '2017-10-24T08:29:24.672Z',
+  dependents: 3321,
+  ts: undefined,
```
Haroenv:
```suggestion
  types: { ts: '@types/babel-core' },
```

src/__tests__/npm.test.js
```diff
@@ -117,7 +117,7 @@ describe('getDownloads()', () => {
     );
 
     // eslint-disable-next-line no-console
-    console.log('downloads', { jest, angular, holmes });
+    // console.log('downloads', { jest, angular, holmes });
```
Haroenv:
```suggestion
    console.log('downloads', { jest, angular, holmes });
```

(easier to update like this when the test get outdated)

Haroenv:
@orta I don't seem to be able to commit my comments, can you do it or give access to the repo? I'd like to try it out live

Haroenv:
hah, faster than my  comment

orta:
👍

karlhorky:
Awesome, merged! 🙌

I guess the next step to get this online is for someone to do a PR to [yarnpkg/website](https://github.com/yarnpkg/website) as per 
@arcanis' [tweet](https://twitter.com/arcanis/status/1139514822481985537)?

And then something for the npm website too?

cc @mxstbr

Haroenv:
Yes, the next step would be to add this to the website since it's now fully  indexed

Haroenv:
https://github.com/yarnpkg/website/blob/master/js/src/lib/Details/Aside.js here probably

Haroenv:
Looks like something went wrong here still, I'll investigate, but the whole index has `ts: null` as the types key

orta:
Doh!

PR #357, opened 2019-07-08
fix(TS): infer definitions correctly

Haroenv:
This fixes multiple things (I was too lazy to make multiple PRs)

1. calculate data based on the package which doesn't end up in the index in formatPkg
2. change the npm endpoint to `registry`

cc @orta

### DIFF
--- a/src/__tests__/__snapshots__/config.test.js.snap
+++ b/src/__tests__/__snapshots__/config.test.js.snap
@@ -179,7 +179,7 @@ Object {
   "maxObjSize": 450000,
   "npmDownloadsEndpoint": "https://api.npmjs.org/downloads",
   "npmRegistryEndpoint": "https://replicate.npmjs.com/registry",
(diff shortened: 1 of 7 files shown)


src/__tests__/__snapshots__/config.test.js.snap
```diff
@@ -179,7 +179,7 @@ Object {
   "maxObjSize": 450000,
   "npmDownloadsEndpoint": "https://api.npmjs.org/downloads",
   "npmRegistryEndpoint": "https://replicate.npmjs.com/registry",
-  "npmRootEndpoint": "https://api.npmjs.org",
+  "npmRootEndpoint": "https://registry.npmjs.org",
```
orta: doh!

Haroenv: I didn't realise the error because it was supposed to check for 404 haha

PR #364, opened 2019-07-25
fix(TS): support scoped packages

Haroenv:
Also changed the schema to make it easier to filter based on type

cc @orta


orta:
Oh yeah, this makes sense!

PR #371, opened 2019-08-02
feat: refacto (part 1)

bodinsamuel:
I would advice reading the 2 commits separately.
Initially wanted to do more but let's not push too much in a single PR.

## Changes
- a8c37e9: add metrics with datadog
- d71c2dd: use async/await syntax everywhere 

## Testing

Launch the 3 different status (bootstrap, replicate, watch), by simply commenting each of the steps.

## Deploy
- Datadog: it will need to be installed with this [tutorial](https://docs.datadoghq.com/agent/basic_agent_usage/heroku/) (never tried though)

## What's next (probably)
- Split code and rename some function (for clarity)
- Improve CHANGELOG 
  JsDelivr returns a list of files with stats, we could use that as a main way to fetch CHANGELOG and rollback to old method if nothing found (some packages are not bundled with it). That could fasten the process.
- Parallelise some stuff
   Saving to algolia and fetching new package for example


src/index.js
```diff
@@ -30,18 +31,29 @@ const { index: bootstrapIndex } = createAlgoliaIndex(c.bootstrapIndexName);
...
   // first we make sure the bootstrap index has the correct settings
+  log.info('💪  Setting up Algolia');
   await setSettings(bootstrapIndex);
+  datadog.timing('main.init_algolia', Date.now() - start);
+
   // then we run the bootstrap
   // after a bootstrap is done, it's moved to main (with settings)
   // if it was already finished, we will set the settings on the main index
+  start = Date.now();
+  log.info('🗻  Bootstraping');
```
Haroenv: tiny nit, but the emoji used for bootstrap is ⛷, replicate 🐌 & watch 🛰

src/index.js
```diff
@@ -30,18 +31,29 @@ const { index: bootstrapIndex } = createAlgoliaIndex(c.bootstrapIndexName);
...
   await bootstrap(await stateManager.check());
+  datadog.timing('main.bootsrap', Date.now() - start);
```
Haroenv: I was considering suggesting `process.hrtime(start)`, but that seems to have a lot more precision we don't need

src/index.js
```diff
@@ -59,35 +71,35 @@ async function setSettings(index) {
...
-      `${emoji} Synced %d/%d changes (%d%), current rate: %d changes/s (%s remaining)`,
-      seq,
-      npmInfo.seq,
-      Math.floor((Math.max(seq, 1) / npmInfo.seq) * 100),
-      Math.round(ratePerSecond),
-      ms(remaining)
-    );
-    loopStart = Date.now();
-  });
+async function infoChange(seq, nbChanges, emoji) {
```
Haroenv: you  changed the other function to logProcess, should this one also be changed to logChange maybe? although they seem extremely similar. Maybe they can even be merged into a single function

src/index.js
```diff
@@ -97,78 +109,95 @@ async function bootstrap(state) {
 
   if (state.seq > 0 && state.bootstrapDone === true) {
     await setSettings(mainIndex);
-    log.info('⛷ Bootstrap: done');
+    log.info('☑️   Bootstrap: done');
```
Haroenv: same emoji remark, I use it to search in the logs sometimes

src/index.js
```diff
@@ -97,78 +109,95 @@ async function bootstrap(state) {
...
-        return saveDocs({ docs: res.rows, index: bootstrapIndex })
-          .then(() =>
-            stateManager.save({
-              bootstrapLastId: newLastId,
-            })
-          )
-          .then(() => infoDocs(res.offset, res.rows.length, '⛷'))
-          .then(() => loop(newLastId));
-      });
+async function loop(lastId) {
```
Haroenv: good call on moving this out, but shouldn't it be called something like "bootstrapLoop" or something similar now? That would avoid confusion on when to use it.

src/index.js
```diff
@@ -97,78 +109,95 @@ async function bootstrap(state) {
...
-          });
+  log.info('-----');
+  log.info(`Total packages   ${totalDocs}`);
+  log.info('-----');
 
-          return moveToProduction();
-        }
+  let still = true;
+  while (still) {
+    still = await loop(state.bootstrapLastId);
```
Haroenv:
```suggestion
    lastProcessedId = await loop(state.bootstrapLastId);
```

A little confusing to mix the return type (boolean) with string, maybe something better can be found?

src/index.js
```diff
@@ -189,29 +218,32 @@ async function replicate({ seq }) {
...
-      saveDocs({ docs, index: mainIndex })
-        .then(() => infoChange(docs[docs.length - 1].seq, 1, '🐌'))
-        .then(() =>
-          stateManager.save({
-            seq: docs[docs.length - 1].seq,
-          })
-        )
-        .then(() => done())
-        .catch(done);
+    const q = cargo(async docs => {
```
Haroenv: did you update async in this PR? I think we are a major version behind, which works a lot better with promises

src/index.js
```diff
@@ -247,36 +279,34 @@ async function watch({ seq }) {
...
+          await stateManager.set({
+            seq: 0,
+            bootstrapDone: false,
+          });
+          process.exit(0); // eslint-disable-line no-process-exit
+        }
+
+        return null;
+      } catch (e) {
+        return e;
```
Haroenv: is this the same as "return done"?

src/jsDelivr.js
```diff
@@ -12,16 +14,31 @@ function formatHits(pkg) {
...
+    const { body: hitsJSON } = await got(c.jsDelivrHitsEndpoint, {
+      json: true,
+    });
+    hits.clear();
+    hitsJSON.forEach(formatHits);
+  } catch (e) {
+    log.error(e);
+
+    // eslint-disable-next-line no-process-exit
+    process.exit(1);
```
Haroenv: I've actually normally not crashed if jsDelivr failed earlier on purpose, because it's not the most important part of data. If it fails, the rest can still continue

src/saveDocs.js
```diff
@@ -4,45 +4,62 @@ import { getDownloads, getDependents } from './npm.js';
...
-    .then(pkgs => index.saveObjects(pkgs))
-    .then(() => log.info('🔍 Found and saved %d packages', rawPkgs.length));
+  let start2 = Date.now();
+  const pkgs = await addMetaData(rawPkgs);
+  datadog.timing('saveDocs.addMetaData', Date.now() - start2);
+
+  start2 = Date.now();
+  index.saveObjects(pkgs);
+  datadog.timing('saveDocs.saveObjects', Date.now() - start2);
+
```
Haroenv: the log "found and saved N packages is gone here now, is it now in index.js?

src/saveDocs.js
```diff
@@ -4,45 +4,62 @@ import { getDownloads, getDependents } from './npm.js';
...
-        ...downloads[index]._searchInternal,
-        ...dependents[index]._searchInternal,
-        ...changelogs[index]._searchInternal,
-        ...hits[index]._searchInternal,
-      },
-    }))
-  );
+  ]);
+
+  return pkgs.map((pkg, index) => ({
```
Haroenv: I think we should add some timings around the whole block here too. I wonder if it could be improved by partialUpdateObject or something like that, or less looping & spreading (constructing the object key by key might be faster?).

src/typescriptSupport.js
```diff
@@ -57,6 +58,11 @@ export async function getTypeScriptSupport(pkg) {
...
  * @param {Array<Package>} pkgs
  */
-export function getTSSupport(pkgs) {
-  return Promise.all(pkgs.map(getTypeScriptSupport));
+export async function getTSSupport(pkgs) {
+  const start = Date.now();
+
+  const all = await Promise.all(pkgs.map(getTypeScriptSupport));
+
+  datadog.timing('changelogs.getTSSupport', Date.now() - start);
```
Haroenv: this isn't for changelogs, but addMetaData

src/index.js
bodinsamuel: For the ref, in the end we can build this kind of dasboard 
Screenshot 2019-08-03 14 39 47

src/index.js
bodinsamuel: fair enough

src/index.js
bodinsamuel: They are similar but not quiet, I initially wanted to merge them but I will wait until I split the code and have a better understanding of the whole process.
I'll rename to something similar for consistency 👌

src/index.js
bodinsamuel: true 👍

src/index.js
bodinsamuel: indeed, I will use `state.bootstrapLastId` for clarity 👍

src/index.js
bodinsamuel: `done` is not defined when using async/await, it should be similar. All my tests seems to proove so, but your tests will be mandatory on this.

src/jsDelivr.js
bodinsamuel: There was no `catch` before so I interpreted it that way. I think it was prone to crash because the error (if any) would bubble up. I'll remove the process.exit 👌

src/saveDocs.js
bodinsamuel: yes, I wanted to have the maximum of logs at same place. 
I should have put a screenshot, but the log looks like that right now for better observability (all naming and log are debatable ☺️ )

Screenshot 2019-08-03 14 56 46

src/typescriptSupport.js
bodinsamuel: oops, thanks

src/index.js
bodinsamuel: done

bodinsamuel:
@Haroenv I answered all your question and let some unresolved where you could still have comments. 
- Upgraded `async`
- Added more metrics
- Fixed a bug in bootstrap
- Fixed emojis
- Fixed naming

MartinKolarik:
>Improve CHANGELOG
JsDelivr returns a list of files with stats, we could use that as a main way to fetch CHANGELOG and rollback to old method if nothing found (some packages are not bundled with it). That could fasten the process.

We also recently discussed with @Haroenv over email the possibility of using this file listing endpoint for typings discovery instead of the current unpkg method, so if the same listing request results could be used for both, the performance improvement should be even bigger.

bodinsamuel:
@MartinKolarik Thanks a lot Martin, did not think about that 🙏 

I'm not sure how to achieve everything in one single call though. 
If I call:
- Stats: https://data.jsdelivr.com/v1/package/npm/algoliasearch/stats I won't have .d.ts/changelog because no ones load them.
- Structure: https://data.jsdelivr.com/v1/package/npm/algoliasearch@3.33.0/tree I won't have stats
- HTML: https://cdn.jsdelivr.net/npm/algoliasearch@3.33.0/ it has everything but I need to parse HTML :'( 

Do you have any advice on this? ☺️ 

Note, we currently download the whole jsdelivr stats in one call https://data.jsdelivr.com/v1/stats/packages/month/all but it seems to fail pretty often and it has a huge memory footprint, I will maybe change that to a call per package, do you think it will be okay? :)

MartinKolarik:
As for stats, I believe the best way is to keep loading them in one call. We added that endpoint specifically for npm-search as it's way more effective than making a request for every package. It should also be stable now - do you have any stats about how often you see it fail?

There is one easy improvement to make though - I just noticed the endpoint returns stats for both npm packages (about 70k records) and GitHub packages (950k records). Since you only use npm packages stats, we could probably add something like https://data.jsdelivr.com/v1/stats/packages/npm/month/all, to reduce the response size by more than 90%.

For typings and changelogs, you would use https://data.jsdelivr.com/v1/package/npm/algoliasearch@3.33.0 (this endpoint also gets faster over time because we now store all listings in a persistent database after generating them on the first request).

bodinsamuel:
> do you have any stats about how often you see it fail?

Yesterday I constantly had 524 coming from cloudflare, I thought it was the endpoint doing that but it does not seems to happen anymore 🙌 

>  to reduce the response size by more than 90%.

That would be awesome 😱 

> For typings and changelogs, you would use https://data.jsdelivr.com/v1/package/npm/algoliasearch@3.33.0 

Thanks a lot, I will put that in place in a next PR 😍

MartinKolarik:
> https://data.jsdelivr.com/v1/stats/packages/npm/month/all

71 MB -> 4 MB

> Thanks a lot, I will put that in place in a next PR 😍

👍 feel free to ping me if you need anything.

bodinsamuel:
trully amazing, thank you !!!

Haroenv:
Note that we fetch changelog from GitHub because sometimes it isn’t inclues in the npm package, and we want to have a fallback there

PR #374, opened 2019-08-04
feat: upgrade packages

bodinsamuel:
Merge all upgrades


.babelrc
```diff
@@ -1,20 +1,17 @@
 {
   "presets": [
     [
-      "env",
+      "@babel/env",
```
Haroenv: awesome

package.json
```diff
@@ -12,8 +12,7 @@
     "test": "jest && eslint .",
     "test:watch": "jest --watchAll --no-watchman",
     "format": "prettier --write **/*.{js,md,json}",
-    "doctoc": "doctoc --notitle --maxlevel 3 README.md",
-    "precommit": "lint-staged"
```
Haroenv: if you removed this, you can also remove the lint-staged config from package.json. I found it useful, but understand it can be slow if you're doing many commits. If I'm sure my commit would pass I pass `-n`. Anyway, removing is fine for me too

Haroenv: Oh, I misread what was going on, I thought you removed it, but it's just moved

bodinsamuel: they changed the API, not sure it works the same though.

PR #379, opened 2019-08-05
chore: move npm code

bodinsamuel:
- Give NPM it's own folder for clarity
- Wrap `db` method that were used in the code, so we can later easily mock them
- Split download code


src/npm/__tests__/index.test.js
```diff
@@ -47,30 +47,42 @@ describe('getDependents()', () => {
...
+      start: expect.any(String),
+      end: expect.any(String),
+      package: '@angular/core',
+    });
+  });
 
-    // real should be 0
-    expect(holmes).toHaveLength(1);
+  it('should download 2 packages and return correct response', async () => {
+    const dl = await api.getDownload('jest,holmes.js');
```
Haroenv: should we not have an api which accepts an array, and keep the handling with adding the extra comma?

Haroenv: Oh I'm seeing now this is a new private import, seems fine

src/npm/index.js
```diff
@@ -153,14 +222,17 @@ export async function getDownloads(pkgs) {
...
-    pkgs.map(() => ({
-      dependents: 0,
-      humanDependents: '0',
-    }))
-  );
+  datadog.timing('npm.getDownloads', Date.now() - start);
+  return all;
 }
+
+export default {
```
Haroenv: I would prefer this to be 
```suggestion
export {
```

and `import * as npm from './npm'` to be able to tree shake it better

Haroenv: oops I just remembered this isn't a library, but still prefer the form I suggest. 

This will still make it easier for us to move functions into a separate file, because we will be able to write `export { nameOfFun } from './xxx'` in this file

src/npm/__tests__/index.test.js
```diff
@@ -128,9 +128,6 @@ describe('getDownloads()', () => {
       pkg.downloadsLast30Days.toString()
     );
 
-    // eslint-disable-next-line no-console
-    console.log('downloads', { jest, angular, holmes });
```
Haroenv: this log was here to be able to update the values easily if the test fails

src/npm/index.js
bodinsamuel: While I agree on the principle, 2 things comes into my mind:
- We usually never move code like that
- It's hard to onboard a codebase where name are too generic. 
  `findAll()` can literally mean anything, with `npm.findAll()` I don't have to go up, find the file to understand the line etc... another solution can be to have the full purpose in the function name but it's really redundant when you are inside the file, e.g: `npmFindAll()`.
  For one line it's obviously not a problem but over a whole codebase it does make a difference. Just my 2 cents on this, if you still want named export I'm fine with it, don't hesitate to request it 👍 ☺️

Haroenv: This is why I'm suggesting `import * as npm from './npm'`, and then doing `npm.findAll`

bodinsamuel: indeed, that would work 👌

src/npm/__tests__/index.test.js
bodinsamuel: Jest already provides a diff if you have anything that does match expectation but fair enough :)

PR #380, opened 2019-08-06
fix: multiple small bugs after #379

bodinsamuel:
- Missing await
- Missing destruct
- Missing .js
- Added log for monitoring purpose

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -155,9 +155,7 @@ async function bootstrapLoop(lastId) {
     options.skip = 1;
   }
 
-  const res = await npm.findAll({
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -156,7 +156,7 @@ async function bootstrapLoop(lastId) {
   }
 
   const res = await npm.findAll({
-    options,
+    ...options,
```
Haroenv: why not just pass options?

bodinsamuel: 🤦‍♂ 
That's probably what I intended to do first

PR #383, opened 2019-08-06
feat(formatPkg): add .js to alternative names

### LINKED ISSUE #217: Idea: drop `.js` endings to provide better relevancy
When searching for "places" we do not find places.js while it's way more popular than `places`
image

image

Haroenv:
fixes #217

1. split out getAlternativeNames
2. avoid duplicates in getAlternativeNames
3. no longer put concatenatedName in _searchInternal (not relied upon)
4. add .js or non-.js name
5. small prettier fix
6. update readme

### DIFF
--- a/src/__tests__/formatPkg.test.js
+++ b/src/__tests__/formatPkg.test.js
@@ -25,10 +25,10 @@ it('keeps .bin intact', () => {
   );
   const formatted = formatPkg(createInstantSearchApp);
   expect(formatted.bin).toMatchInlineSnapshot(`
-Object {
-  "create-instantsearch-app": "src/cli/index.js",
-}
-`);
+        Object {
+          "create-instantsearch-app": "src/cli/index.js",
+        }
+    `);
 });
 
 it('truncates long readmes', () => {
@@ -223,7 +223,7 @@ describe('adds TypeScript information', () => {
   });
 });
 
-describe('test getRepositoryInfo', () => {
+describe('getRepositoryInfo', () => {
   const getRepositoryInfo = formatPkg.__RewireAPI__.__get__(
     'getRepositoryInfo'
   );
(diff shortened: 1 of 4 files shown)


src/__tests__/formatPkg.test.js
```diff
@@ -377,3 +377,67 @@ describe('test getRepositoryInfo', () => {
...
   });
 });
+
+describe('alternative names', () => {
+  test('name not yet ending in .js', () => {
+    const original = {
+      name: 'places',
+      lastPublisher: { name: 'unknown' },
+    };
+    expect(formatPkg(original)._searchInternal.alternativeNames)
```
bodinsamuel: If you are specifically testing `getAlternativeNames` you could use it directly, that way the test would be clear about the expectation. You have already have snapshot in place to be sure that this function is already called in the main method.

src/__tests__/formatPkg.test.js
```diff
@@ -377,3 +377,67 @@ describe('test getRepositoryInfo', () => {
...
+  test('scoped package', () => {
+    const original = {
+      name: '@algolia/places.js',
+      lastPublisher: { name: 'unknown' },
+    };
+    expect(formatPkg(original)._searchInternal.alternativeNames)
+      .toMatchInlineSnapshot(`
+            Array [
+              "algoliaplacesjs",
+              " algolia places js",
```
bodinsamuel: maybe a `.trim()` would clean that

Haroenv: yes, however it isn't really an issue for the engine, so I left it as-is

src/__tests__/formatPkg.test.js
Haroenv: good point, but would leave it for a major reorganise, since these functions aren't exported :)

PR #384, opened 2019-08-06
feat(jsDelivr): move code, add tests, preload data correctly

bodinsamuel:
- Move code like npm, for clarity
- Add unit tests for all functions. It's not mocked but at least it tests the expectations. 
  Thanks to @MartinKolarik it loads pretty fast now
- JsDelivr hits were only loaded in Bootstrap mode which mean we were not pushing hits in watch mode
- Add `getFilesList` and `getAllFilesList` that we will use later to compute TS and Changelog.


src/jsDelivr/index.js
```diff
@@ -0,0 +1,94 @@
+import got from 'got';
+import log from '../log.js';
+import config from '../config.js';
+import datadog from '../datadog.js';
+
+const hits = new Map();
+
+function formatHits(pkg) {
```
Haroenv: would now rename this & make it inline, so it would become just the next suggestion, and removing this function

src/jsDelivr/index.js
```diff
@@ -0,0 +1,94 @@
...
+async function loadHits() {
+  const start = Date.now();
+  log.info('📦  Loading hits from jsDelivr');
+
+  try {
+    const { body: hitsJSON } = await got(config.jsDelivrHitsEndpoint, {
+      json: true,
+    });
+    hits.clear();
+    hitsJSON.forEach(formatHits);
```
Haroenv:
```suggestion
    hitsJSON.forEach(pkg => {
      hits.set(pkg.name, pkg.hits);
    });
```

WDYT?

src/jsDelivr/index.js
```diff
@@ -0,0 +1,94 @@
...
+  const start = Date.now();
+  if (!pkg.name || !pkg.name.includes('@')) {
+    throw new Error(
+      `Package name should contain a version number: ${pkg.name}`
+    );
+  }
+
+  let files;
+  try {
+    files = await got(`${config.jsDelivrPackageEndpoint}/${pkg.name}/flat`, {
```
Haroenv: we can put the return inside this try to avoid having a `let`. Not really a big deal though.

Annoying thing is that eslint might complain about "inconsistent return", but I'm not sure it will

src/jsDelivr/__test__/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,183 @@
...
+      "time": "1985-10-26T08:15:00.000Z",
+    },
+    Object {
+      "hash": "kt2uoTK/NmyQe2OUiNwpdwxV4RxgS2gW9rEgOtj+lZU=",
+      "name": "/tsconfig.tsbuildinfo",
+      "size": 220798,
+      "time": "1985-10-26T08:15:00.000Z",
+    },
+  ],
+  Object {},
```
Haroenv: should this not be an empty array, instead of empty object?

src/__tests__/__snapshots__/config.test.js.snap
```diff
@@ -177,6 +177,7 @@ Object {
     },
   ],
   "jsDelivrHitsEndpoint": "https://data.jsdelivr.com/v1/stats/packages/npm/month/all",
+  "jsDelivrPackageEndpoint": "https://data.jsdelivr.com/v1/package/npm",
```
Haroenv: let's remove this snapshot and just assert nothing changed in the test, instead of needing to update this every time. It seems fairly useless

Haroenv: (in a separate PR)

Haroenv:
> JsDelivr hits were only loaded in Bootstrap mode which mean we were not pushing hits in watch mode

good point, however, bootstrap is always ran on "machine init". We need to be careful to actually request fresh results from time to time, not just on startup. That can be done in a separate PR though!

Haroenv:
Note that `getFilesList` might not be enough to replace the ad-hoc requesting from unpkg, because it will also be able to request deep files. However, this would make the "general case" a lot faster, and we can still rely on the network for deeper files

src/jsDelivr/__test__/__snapshots__/index.test.js.snap
bodinsamuel: good call

src/jsDelivr/index.js
bodinsamuel: 👍

src/jsDelivr/index.js
bodinsamuel: It was more to have a single metric call that way we don't forget to count a branch in the metric.

src/__tests__/__snapshots__/config.test.js.snap
bodinsamuel: yes 👌

bodinsamuel:
> good point, however, bootstrap is always ran on "machine init".

yes but as the process crash often, as some point we no longer have data.

>  fresh results from time to time, not just on startup

good idea, we could have an auto refresh in a setInterval or some kind of `lastRefresh` date stored somewhere.

> Note that getFilesList might not be enough

Yes it will not be enough, but if we can optimize the top 100k, it's 10% easily won

MartinKolarik:
@Haroenv what do you mean by "deeper files"?

Haroenv:
Retrieving a file not at the root of a package would require further api calls. I think it’s a fine trade off to avoid loading the whole file tree for every package which doesn’t need it

MartinKolarik:
But our API does return the whole file tree.

bodinsamuel:
@Haroenv I'm specifically requiring `/flat` to make iteration easier but it's the full tree 👍

bodinsamuel:
@Haroenv Do you have anything to add? ☺️

src/jsDelivr/index.js
```diff
@@ -0,0 +1,95 @@
...
+  return files;
+}
+
+/**
+ * Get one package files list
+ * @param {object} pkg
+ */
+async function getFilesList(pkg) {
+  const start = Date.now();
+  if (!pkg.name || !pkg.name.includes('@')) {
```
Haroenv: what about using `pkg.name` & `pkg.version`? Can be changed once we actually use this method though

bodinsamuel: good point, I assumed the param was computed but we usually pass a Pkg object that contains everything.

PR #385, opened 2019-08-06
feat: move algolia

bodinsamuel:
- Split again code to remove from index.js and clarify
- Remove global variables so we prepare the field for more mocking and code splitting
-


src/algolia/index.js
```diff
@@ -0,0 +1,62 @@
...
+ * @param {object} config
+ */
+async function prepare(config) {
+  // Get main index and boostrap algolia client
+  const { index: mainIndex, client } = createClient(
+    config.appId,
+    config.apiKey,
+    config.indexName
+  );
+  const { index: bootstrapIndex } = createClient(
```
Haroenv: looking at this code, it might make a lot more sense to simply call "algoliasearch()" twice now, and put the error here, instead of the wrapper. WDYT?

src/index.js
```diff
@@ -14,9 +15,7 @@ log.info('🗿 npm ↔️ Algolia replication starts ⛷ 🐌 🛰');
 
 let loopStart = Date.now();
 
-const { index: mainIndex, client } = createAlgoliaIndex(config.indexName);
-const { index: bootstrapIndex } = createAlgoliaIndex(config.bootstrapIndexName);
-const stateManager = createStateManager(mainIndex);
+let stateManager;
```
Haroenv: is this just to have the variable global? I'd rather pass it around than have a mutable (maybe undefined) variable :)

bodinsamuel: It's just to have the variable global for the moment, I'll change that in an other PR. 
I can do it now if you want

Haroenv: would make sense here to me :)

bodinsamuel: Did it, plus added some test 🙆‍♂

bodinsamuel: 3f6dfb1

src/algolia/index.js
bodinsamuel: I moved the error sooner, I'll let the wrapper for algoliasearch It's cleaner for var declaration

src/index.js
```diff
@@ -35,31 +33,26 @@ async function main() {
...
   // if it was already finished, we will set the settings on the main index
   start = Date.now();
   log.info('⛷   Bootstraping');
-  await bootstrap(
-    await stateManager.check(),
-    algoliaClient,
-    mainIndex,
-    bootstrapIndex
-  );
+  await bootstrap(stateManager, algoliaClient, mainIndex, bootstrapIndex);
```
Haroenv: seems inconsistent between bootstrap which checks within the function and replicate which checks outside?

bodinsamuel: indeed

bodinsamuel:
Do you have anything to add? ☺️

PR #386, opened 2019-08-06
feat(data): add flagging of type=module

Haroenv:
I've looked around, and it seems like this is a sufficient heuristic to detect esm. I have explicitly left the most common case as unknown, so we can later find out which type those are.

References:
- https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field
- https://github.com/rollup/rollup/wiki/pkg.module

There's also "jsnext:main", but that doesn't seem to be used anymore

cc @fredkschott, is this logic correct?


README.md
```diff
@@ -121,6 +121,7 @@ For every single NPM package, we create a record in the Algolia index. The resul
     ts: 'definitely-typed', // definitely-typed | included | false
     definitelyTyped: '@types/babel__core',
   },
+  moduleType: 'cjs', // esm | cjs | unknown
```
bodinsamuel: I would make that an array, some package build for esm and cjs and browser and umd...

src/formatPkg.js
```diff
@@ -431,3 +433,18 @@ function getAlternativeNames(name) {
...
+    typeof cleaned.module === 'string' ||
+    cleaned.type === 'module' ||
+    main.endsWith('.mjs')
+  ) {
+    return 'esm';
+  }
+  if (cleaned.type === 'commonjs' || main.endsWith('.cjs')) {
+    return 'cjs';
+  }
+  return 'unknown';
```
bodinsamuel: That is good iteration to begin with. Like I said it's may be a bit too simplistic but it can be improved later.

README.md
Haroenv: Good idea, umd or browser build is too hard to distinguish though. Can check for the browser field, but is only necessary if package is needing  environment-specific code. Just cjs & esm where I’m 100% sure for now

src/__tests__/__snapshots__/config.test.js.snap
```diff
@@ -96,6 +96,7 @@ Object {
       "searchable(owner.name)",
       "deprecated",
       "types.ts",
+      "moduleType",
```
Haroenv:
```suggestion
      "moduleTypes",
```

src/config.js
```diff
@@ -35,6 +35,7 @@ const defaultConfig = {
       'searchable(owner.name)',
       'deprecated',
       'types.ts',
+      'moduleType',
```
Haroenv:
```suggestion
      'moduleTypes',
```

PR #387, opened 2019-08-08
fix(algolia): missing config param

bodinsamuel:
Well 🥇


Haroenv:
oopsie, thanks for fixing so fast!

PR #390, opened 2019-08-08
feat: add Sentry

bodinsamuel:
Add sentry for error reporting.


src/utils/sentry.js
```diff
@@ -0,0 +1,25 @@
+import Sentry from '@sentry/node';
+
+import log from './../log.js';
```
Haroenv:
```suggestion
import log from '../log.js';
```

src/index.js
```diff
@@ -324,7 +326,8 @@ async function watch(stateManager, mainIndex) {
   });
 }
 
-function error(err) {
-  log.error(err);
+async function error(err) {
+  sentry.report(err);
+  await wait(5000);
   process.exit(1); // eslint-disable-line no-process-exit
```
Haroenv: if we just rethrow the error, will sentry still catch it, or do we need the timeout?

src/utils/sentry.js
bodinsamuel: I have this habits, not sure why :D

src/index.js
bodinsamuel: The timeout is needed (I will add a comment) if I don't wait the `process.exit` come too soon and the error don't have time to reach their API (I guess)

PR #391, opened 2019-08-08
feat(sentry): wait for the right amount of time.

Haroenv:
I didn't test this, just found it in the docs: https://docs.sentry.io/error-reporting/configuration/draining/?platform=node


src/index.js
```diff
@@ -328,9 +327,6 @@ async function watch(stateManager, mainIndex) {
 
 async function error(err) {
   sentry.report(err);
```
Haroenv: we can pass more info here wrt the json problem, but I couldn't find where to get the original json

PR #393, opened 2019-08-09
feat(index): try being resilient on "changes"

Haroenv:
There doesn't seem to be a builtin way of retrying on .changes pouchdb (https://github.com/pouchdb/pouchdb/issues/7209)

I tried this out, but I have not been able to make this fail on my local machine, because I don't manage to set up a MITM where i can fake a wrong response.

Only  approve this if you have tried it out and see no issues

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -194,7 +194,7 @@ async function moveToProduction(stateManager, algoliaClient) {
   await stateManager.save(currentState);
 }
 
-async function replicate(stateManager, mainIndex) {
+async function replicate(stateManager, mainIndex, retry = 0, retryError) {
   const { seq } = await stateManager.get();
   log.info(
     '🐌   Replicate: Asking for %d changes since sequence %d',
@@ -209,11 +209,18 @@ async function replicate(stateManager, mainIndex) {
   const { seq: npmSeqToReach } = await npm.getInfo();
   let npmSeqReached = false;
 
+  if (retry > config.replicateRetry) {
+    return Promise.reject(
+      new Error('replicate has failed > 3 times'),
+      retryError
+    );
+  }
+
   return new Promise((resolve, reject) => {
     const listener = npm.listenToChanges({
       since: seq,
-      batch_size: config.replicateConcurrency, // eslint-disable-line camelcase
       live: true,
+      batch_size: config.replicateConcurrency, // eslint-disable-line camelcase
       return_docs: false, // eslint-disable-line camelcase
     });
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -313,20 +333,27 @@ async function watch(stateManager, mainIndex) {
...
       });
     });
-    listener.on('error', reject);
+
+    listener.on('error', err => {
+      sentry.report(err, {
+        retry: true,
+      });
+      log.info('🛰  Retrying watch', retry + 1);
+      watch(stateManager, mainIndex, retry + 1, err);
```
bodinsamuel: need async/await probably

src/index.js
```diff
@@ -313,20 +333,27 @@ async function watch(stateManager, mainIndex) {
...
+    });
   });
 }
 
-async function error(err) {
-  sentry.report(err);
+async function error(err, extra) {
+  sentry.report(err, extra);
```
bodinsamuel: This function is never called directly, it can not have extra.

src/index.js
```diff
@@ -250,9 +257,18 @@ async function replicate(stateManager, mainIndex) {
...
+    let lastError = undefined;
+    listener.on('error', err => {
+      lastError = err;
+      listener.cancel();
+    });
 
     changesConsumer.drain(() => {
+      if (lastError) {
+        log.info('🐌  Retrying replicate', retry + 1);
+        replicate(stateManager, mainIndex, retry + 1, lastError);
```
bodinsamuel: need async/await probably

src/index.js
```diff
@@ -250,9 +257,18 @@ async function replicate(stateManager, mainIndex) {
         listener.cancel();
       }
     });
-    listener.on('error', reject);
+    let lastError = undefined;
+    listener.on('error', err => {
+      lastError = err;
+      listener.cancel();
```
bodinsamuel: On cancel does the `.drain()` is called ?

src/index.js
Haroenv: Good call, needs a return

src/index.js
Haroenv: We want to early cancel if eg 2/10 fails, we don’t want to continue with the other ones, or else it’ll skip one & be complicated

src/index.js
Haroenv: The reject should bubble all the way up to main, no?

bodinsamuel: true, I didn't saw your second params on the reject. Not sure it work though `reject` only accept one param

PR #394, opened 2019-08-09
feat: dockerify + kubernetesable

bodinsamuel:
Just a test for the moment,
I'm testing the script in a machine in our private google cloud kubernetes cluster (1cpu/2GO ram) to see how it goes.


bodinsamuel:
JSON stream error happened at the exact same second as the prod one 👌 
Still don't know why but at least it's consistant

kubernetes.yml
```diff
@@ -0,0 +1,39 @@
...
+            - yarn
+            - start
+          resources:
+            requests:
+              cpu: '1'
+              memory: '2000Mi'
+
+          env:
+            - name: appId
+              value: 'KECOEVPMGH'
```
Haroenv: this should probably no the hardcoded right?

bodinsamuel:
It launches `yarn start` too, it's just that it's bigger, managed differently and cheaper.

Haroenv:
makes sense, it doesn't really seem much more complicated, and if it's faster & cheaper then it's good :)

bodinsamuel:
No longer needed

PR #395, opened 2019-08-10
fix(formatPkg): cleaned main can be an array

bodinsamuel:


Haroenv:
Can we add a test for that?

PR #402, opened 2019-08-12
process package in the right order

bodinsamuel:
I knew it was too easy.
Basically reintroduce the queue that I though a bit useless but it was not, added comment to reflect on this

### DIFF
--- a/src/watch.js
+++ b/src/watch.js
@@ -1,7 +1,9 @@
+import ms from 'ms';
+import queue from 'async/queue.js';
+
 import config from './config.js';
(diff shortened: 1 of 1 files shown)


Haroenv:
is this not necessary in catchup too?

bodinsamuel:
That was my reflection right now, this could definetely happen at catchup time. Already broken before, still broken now :'( 
If I put the concurrency to 1 in the catchup, this will be a nightmare to actually catchup. Maybe we need to have code some "intelligence" here

PR #412, opened 2019-08-19
fix: kill process regurlarly, for cache and bootstrap

bodinsamuel:
I recent refactos I removed the logic to kill the process in the Watch.


src/index.js
```diff
@@ -11,13 +11,25 @@ import * as watch from './watch.js';
...
  *   - Bootstrap: will index the whole list of packages (if needed)
  *   - Watch    : will process update in real time
  */
 async function main() {
   const start = Date.now();
+
+  // We schedule to kill the process:
+  //  - reset cache
+  //  - maybe retrigger bootstrap
+  setInterval(() => {
```
Haroenv:
```suggestion
  setTimeout(() => {
```

Does this have an impact on perf, I've heard other times long-running timers can take up much more memory than expected, or will it be fine?

bodinsamuel: I tried to find info about this, I think the way nodejs is managing the even loop is making it trivial (or at least consistant) wether it's long or not.

bodinsamuel: NB: you can have up to 25days (32bit) of timeout, so they probably handle that fine

bodinsamuel:
Actually we can merge because I remove the code to reset the bootstrap with `timeToRedoBootstrap`
I'll reintroduce it after everything is done 👍

PR #413, opened 2019-08-23
feat: parallel fetch

bodinsamuel:
- add `packagesFetcher` 
   It will prefetch at start and continuously fill a storage will we do `saveDocs()`


src/bootstrap.js
```diff
@@ -32,27 +34,52 @@ async function run(stateManager, algoliaClient, mainIndex, bootstrapIndex) {
...
+  // Firsy sync packagesFetcher
+  Promise.all([
+    await packagesFetcher.syncTotalWithNPM(),
+    await packagesFetcher.syncOffset(),
+  ]);
+
   log.info('-----');
-  log.info(`Total packages   ${totalDocs}`);
+  log.info(`Total packages   ${packagesFetcher.total}`);
+  log.info(`Starting offset   ${packagesFetcher.nextOffset}`);
```
Haroenv: what's the difference between nextOffset & nextKey? number & name?

src/bootstrap.js
```diff
@@ -129,6 +165,29 @@ async function logProgress(offset, nbDocs) {
     Math.round(ratePerSecond),
     ms(((totalDocs - offset - nbDocs) / ratePerSecond) * 1000)
   );
+
+  // ------- Last minutes
+  const thisMinute = new Date().getMinutes();
```
Haroenv: not sure if I understand it correctly, but this logic might be wrong across day limits? Might be better to work with unix seconds/ ms

Haroenv: if too tricky, I'm fine with using `ms`  or date-fns for this

src/datadog.js
```diff
@@ -4,9 +4,10 @@ import log from './log.js';
 const env = process.env.NODE_ENV === 'production' ? 'prod' : 'dev';
 
 const client = new StatsD({
-  host: process.env.DOGSTATSD_HOST || 'localhost',
+  host: process.env.DOGSTATSD_HOST,
   port: 8125,
   prefix: 'alg.npmsearch.',
+  mock: !process.env.DOGSTATSD_HOST,
```
Haroenv: good find

src/npm/packagesFetcher.js
```diff
@@ -0,0 +1,148 @@
...
+      }
+
+      while (this.storage.length <= 0 || !this.isFinished) {
+        log.warn('🥴  We process packages faster than we prefetch them');
+        await this.prefetch();
+        await wait(5000);
+      }
+    }
+
+    const packages = this.storage.splice(0, this.limit);
```
Haroenv: good call on removing them from the array, mutating will indeed likely be the fastest here

.babelrc
```diff
@@ -1,12 +1,20 @@
...
-    ],
+        "targets": {
+          "node": "current"
+        }
+      }
+    ]
   ],
+  "plugins": [
+    [
+      "@babel/plugin-proposal-class-properties",
```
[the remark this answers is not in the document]
bodinsamuel: It's just needed for jest, to enable the use of properties in packagesFetcher

src/npm/packagesFetcher.js
```diff
@@ -0,0 +1,148 @@
+import log from './../log.js';
+import wait from './../utils/wait.js';
+
+import * as api from './index.js';
+
+class PackagesFetcher {
+  limit = 100;
+  max = 1000;
+  total = 0;
+  storage = [];
```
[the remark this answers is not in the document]
bodinsamuel: I guess it can but we are preloading max 1000 and we saw in heroku that we are on average at 200mb. The max can be tweaked to be more respectful of the limitation later ☺️

src/bootstrap.js
bodinsamuel: key = cursor to paginate
offset = position in the list

We can only paginate with the key, and the offset is only here to say "we are at <offset>/10000000"

src/bootstrap.js
bodinsamuel: `.getMinutes()` returns 0-59.

`thisMinute` is stored on a array `lastMinutes`, like this

```javascript
lastMinutes = [
  [ <minute>, <count> ]
];
```

this enable us to reuse the key "minute" to and have only N entry for N minutes. where N = 10.
**E.g:**
```javascript
lastMinutes = [
  [ 7, <count> ]
  [ 8, <count> ]
  [ 9, <count> ]
  [ 10, <count> ]
  [ 11, <count> ]
  [...]
];
```

at line 171 we shift the array if the length is > higher than N (10), we can not have more than 10 entries so there is never an overlap between hours or days. This we way have rolling stats over the last 10minutes easily.

src/npm/packagesFetcher.js
bodinsamuel: actually I'm still hesitating with having "named page" in a object 
`{ '0-100': [...], '100-200': [...] }` or with the nextKey `{ '@a/a': [...], '@a/ba': [...] }`
this way we could fetch multiples pages in parallel and even process multiple pages in parallel

.babelrc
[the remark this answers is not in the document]
Haroenv: I know, but I don't see properties in that class :)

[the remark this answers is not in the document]
bodinsamuel:
```javaascript
limit = 100;
max = 1000;
total = 0;
storage = [];
[...]
```

bodinsamuel:
It has been ported to #656

PR #516, opened 2020-03-05
check-changelog-on-unpkg-also

Noviny:
Really quick stab at fetching changelog files from either npm or unpkg.

Happy to act on any feedback, or if there's any additional work that needs to be done to make this what we want.

I have not run this code.

### DIFF
--- a/src/changelog.js
+++ b/src/changelog.js
@@ -2,7 +2,7 @@ import got from 'got';
 import race from 'promise-rat-race';
 
 import datadog from './datadog.js';
-import { gotUnpkg } from './unpkg.js';
+import config from './config.js';
 
 export const baseUrlMap = new Map([
   [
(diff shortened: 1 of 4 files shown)


src/changelog.js
```diff
@@ -2,6 +2,7 @@ import got from 'got';
 import race from 'promise-rat-race';
 
 import datadog from './datadog.js';
+import { gotUnpkg } from './unpkg';
```
Haroenv:
```suggestion
import { gotUnpkg } from './unpkg.js';
```

src/unpkg.js
```diff
@@ -15,3 +15,10 @@ export async function fileExistsInUnpkg(pkg, version, path) {
     return false;
   }
 }
+
+export async function gotUnpkg(pkg, version, path) {
```
Haroenv:
```suggestion
export function gotUnpkg(pkg, version, path) {
```

Noviny:
I fixed the linting (I like the function being async so added an await to the return)

I added it all to one race, but another option is to let repo resolve, and only if it can't be resolved do we go looking at unpkg (make it actually sequential). Is there a good way to test the impact of this?

Notes from just trying to run this locally to verify the changes - there's a dev command listed in the docs, but no dev command in the `package.json`. On start, I was told my API key was not valid, which, fair. If you can send me an API key for testing, I have a few packages I could check that would show me this was working.

Haroenv:
sent you credentials in DM

Noviny:
Turns out I probably didn't need credentials 🤷‍♀️

Wrote a test for the changelog fetching. This immediately made the direction of what I wanted to do clearer.

Effectively, there are two arrays of possible changelog paths to check, unpk paths and repository paths.

I pulled out the return logic into its own function. Anywhere we were returning early, it now just runs the race on the unpkg paths.

If the repository info is valid, it runs on both repo and unpkg paths.

src/changelog.js
```diff
@@ -64,14 +56,35 @@ async function getChangelog({ repository }) {
...
+
+  const user = repository.user || '';
+  const project = repository.project || '';
+  const host = repository.host || '';
+  if (user.length < 1 || project.length < 1) {
+    return raceFromPaths(unpkgFiles);
+  }
+
+  // Check if we know how to handle this host
+  if (!baseUrlMap.has(host)) {
```
Haroenv: we can put unpkg in the base url map & keep it as-is maybe?

Noviny: I thought about putting it in the baseUrlMap, but it doesn't quite work in there - it would require passing `name` and `version` into the `baseUrlMap` which doesn't make sense for anything else.

What's the thing you want to change about the code as-is?

Haroenv: makes sense, let's keep it like in your OR

Haroenv:
Trying this out on prod now

PR #520, opened 2020-03-11
handle bb redirect for private repo

Noviny:
Found this fun little catch in fetching info from BB. Feel free to DM me if you want a URL / package to test this on.

Second approach to this would be to somehow flag this in the `baseUrlMap` by complicating it to allow including a callback to perform additional checks, lmk what you think.

### DIFF
--- a/src/__tests__/changelog.test.js
+++ b/src/__tests__/changelog.test.js
@@ -11,7 +11,7 @@ jest.mock('got', () => {
 
   return url =>
     gotSnapshotUrls.has(url)
-      ? Promise.resolve({ url })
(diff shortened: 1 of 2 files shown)


Haroenv:
seems like this makes the other tests fail, is it a bug in the tests or the code?

Noviny:
I fixed this from sort of both angles. 1) the test's mock didn't have `redirectUrls`, but 2) I've made sure we handle cases where they don't exist (they should always exist)

Let me know if you want me to remove the extra check bits in favour of trusting got's types.

Haroenv:
looks good, thanks for your contribution!

PR #592, opened 2021-06-03
fix(dev): upgrade env

bodinsamuel:
- Update packages
- Setup circleci
- Reup travis for the moment
- Upgrade node and yarn
- Upgrade eslint and --fix

image


.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
Haroenv, line 1: should this be a .gitkeep file instead?

package.json
```diff
@@ -69,7 +64,7 @@
     "eslint-plugin-jsdoc": "35.1.2",
     "eslint-plugin-prettier": "3.4.0",
     "eslint-plugin-react": "7.24.0",
-    "husky": "6.0.0",
+    "husky": "^6.0.0",
```
Haroenv, line 67: other dependencies are installed exactly (no caret)

.nvmrc
```diff
@@ -1 +1 @@
-12.10.0
+14.16.1
```
Haroenv, line 1: we can remove --experimental-modules now

Haroenv, line 1: why not migrated to v16?

.husky/.gitignore
bodinsamuel: apparently not (this is the automated install from husky)

Haroenv (commented):
👍

.nvmrc
bodinsamuel: it's not yet LTS, it's in october

PR #595, opened 2021-06-04
fix(dev): upgrade env /2

bodinsamuel:
- Upgrade some others packages (runtime this time)
- Fix got upgrade and use keep alive agent 
- Reduce bootstrap amount (100 does not even pass on my computer)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@sentry/node": "5.30.0",
+    "@sentry/node": "6.5.1",
     "algoliasearch": "3.35.1",
     "async": "3.2.0",
     "babel-cli": "6.26.0",
@@ -31,13 +31,13 @@
     "babel-preset-stage-2": "6.24.1",
     "bunyan": "1.8.15",
     "bunyan-debug-stream": "2.0.0",
-    "dotenv": "8.2.0",
+    "dotenv": "10.0.0",
     "dtrace-provider": "0.8.8",
     "escape-html": "1.0.3",
-    "got": "9.6.0",
+    "got": "11.8.2",
     "gravatar-url": "3.1.0",
     "hosted-git-info": "2.7.1",
-    "hot-shots": "6.8.7",
+    "hot-shots": "8.3.2",
     "lodash": "4.17.21",
     "ms": "2.1.3",
     "nano": "8.2.3",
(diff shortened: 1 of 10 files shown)


src/utils/request.js
```diff
@@ -0,0 +1,25 @@
+import http from 'http';
+import https from 'https';
+
+import got from 'got';
+
+const options = {
+  keepAlive: true,
+  timeout: 60000,
```
Haroenv, line 8:
```suggestion
  timeout: 60_000,
```

package.json
```diff
@@ -23,21 +23,21 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@sentry/node": "5.30.0",
-    "algoliasearch": "3.35.1",
+    "@sentry/node": "6.5.1",
+    "algoliasearch": "4.9.1",
```
Haroenv, line 27: aren't there breaking changes in the algolia usage?

src/utils/request.js
bodinsamuel: this should be an eslint rule 😏

package.json
bodinsamuel: well you are better suited to answer to that, but the basic client creation and save objects seems to work fine.
Note I tested locally with bootstrap with 100 Packages.

src/utils/request.js
bodinsamuel: actually eslint does not liky

package.json
Haroenv: If it was typescript we could find more of the errors, but for example:

```
async function putDefaultSettings(index, config) {
  await index.setSettings(config.indexSettings);

  await index.batchSynonyms(config.indexSynonyms, {
    replaceExistingSynonyms: true,
  });
  const { taskID } = await index.batchRules(config.indexRules, {
    replaceExistingRules: true,
  });

  await index.waitTask(taskID);
}
```

must become:

```
  await index.setSettings(config.indexSettings);

  await index.saveSynonyms(config.indexSynonyms, {
    replaceExistingSynonyms: true,
  });
  const { taskID } = await index.saveRules(config.indexRules, {
    replaceExistingRules: true,
  });

  await index.waitTask(taskID);
}
```

bodinsamuel: ah good catch, this one is executed only if the original index does not exists. I missed it 👌🏻 
in the next PRs I'll try to create integration test, a lot of things are untested right now, too easy to make mistake (and I do a lot of them ahah)

bodinsamuel: Screen Shot 2021-06-04 at 17 21 09

PR #623, opened 2021-06-20
fix(ts): migrate watch.js

bodinsamuel:
The most complexe one. 
I had to change the logic to fix the types (a good thing but introduces more changes than what I wanted).

main issues:
- Catchup was processing multiple updates in parallel but it could have an update and a delete for the same package, 
which could explain one of the issue we have with deleted packages still in the index.
- Repo provided by NicePackage was not as "clean" as we wanted, realised that while testing with real update.
- We were filtering out `deleted` packages, in `changesConsumer` (Line 127, 131) effectively never deleting packages.

### DIFF
--- a/src/__tests__/formatPkg.test.ts
+++ b/src/__tests__/formatPkg.test.ts
@@ -155,11 +155,11 @@ describe('adds vue-cli plugins', () => {
     name: '@dogs/vue-cli-plugin-dogs',
   };
 
-  const formattedDogs = formatPkg(pkg);
-  const formattedUnofficialDogs = formatPkg(unofficialDogs);
-  const formattedScopedDogs = formatPkg(scopedDogs);
-
   it('should format correctly', () => {
+    const formattedDogs = formatPkg(pkg);
+    const formattedUnofficialDogs = formatPkg(unofficialDogs);
+    const formattedScopedDogs = formatPkg(scopedDogs);
+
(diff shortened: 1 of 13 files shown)


src/__tests__/formatPkg.test.ts
```diff
@@ -155,11 +155,11 @@ describe('adds vue-cli plugins', () => {
     name: '@dogs/vue-cli-plugin-dogs',
   };
 
-  const formattedDogs = formatPkg(pkg);
```
Haroenv, line 158: used at line 182, is that fixed in a different PR?

bodinsamuel: Line 182 is a different describe, or maybe I'm miss reading

Haroenv: it's a different describe, but it uses the same variables, right?

bodinsamuel: I'm not sure I get the issue. I'v moved the 3 lines to not be in a describe but in the `test` so that it's not executed every time when you filter some test

Haroenv: my bad, I totally misread it, thought it was in a beforeEach or something first

bodinsamuel: ah, no worries ☺️

PR #626, opened 2021-06-20
feat: full TS migration

bodinsamuel:
If you want to review:
I suggest to look at each commit separately, each one is 1 PR and 1 file.

I managed to introduce very few logic changes, the biggest one is in #623

---

Currently bootstrapping on my personal Algolia Account to test everything out.


src/utils/request.ts
```diff
@@ -14,8 +15,12 @@ const options = {
 export const httpAgent = new http.Agent(options);
 export const httpsAgent = new https.Agent(options);
 
-export async function request(url, opts) {
-  return await got(url, {
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
Haroenv, line 18: not sure if we need this lint rule anywhere

src/watch.ts
```diff
@@ -0,0 +1,237 @@
...
+ * @param mainIndex - Algolia index manager.
+ */
+async function watch(stateManager: StateManager): Promise<true> {
+  const { seq } = await stateManager.get();
+
+  const listener = npm.listenToChanges({
+    since: String(seq),
+    include_docs: false,
+    heartbeat: 30 * 1000,
+    // limit: 100,
```
Haroenv, line 127: does this limit not come from config?

Haroenv (commented):
these are some great changes, thanks @bodinsamuel !

bodinsamuel:
Still indexing (obv), no issue so far:
[progress] 93907/1646400

bodinsamuel:
Still indexing 😴
[progress] 219214/1647440

Haroenv:
Are you indexing on a local machine?

bodinsamuel:
Heroku free plan :D

bodinsamuel:
@material/auto-init (and packages just after) does not fit in memory.

This package has 1027 versions at the time of writing which could be problematic. And can happen to any packages actually.
Screenshot 2021-06-22 at 15 27 38

Haroenv:
At which point does it run out of memory, at parsing of the json already? could we eg. spawn workers or something to avoid the memory load on the main thread, or does it not work like that @bodinsamuel?

bodinsamuel:
> At which point does it run out of memory, at parsing of the json already? 

Hard to say, looking at the fast increase it most likely happened while receiving the response yes. (so outside our control)

> could we eg. spawn workers or something to avoid the memory load on the main thread, or does it not work like that @bodinsamuel?

It's not like CPU, you can not consume more than what you have. Heroku is being super nice by saying "quota vastly exceeded" suggesting that we actually run in a bigger VM that what we pay for, but in a stricter env it would just have crashed.

Unless there is a way to get the package without all versions someway, I'm not even sure we have proper solution for this, the NPM API is pretty dumb.

Haroenv:
In our case we indeed don't need access to all data in the packument, maybe a proxy to simplify the object could work (that could use a different json parser that doesn't put things that aren't needed in memory, like sanity's groq)

bodinsamuel:
Still indexing 😬 
[progress] 1390642/1658725

bodinsamuel:
[progress] 1490841/1659409 docs (89%)

Haroenv (commented):
looks good, although I left a couple comments last time that aren't yet addressed

.eslintrc
```diff
@@ -1,20 +1,24 @@
...
       {
-        "ignorePackages": true
+        "ignorePackages": true,
+        "pattern": {
+          "js": "always",
+          "ts": "never"
+        }
       }
     ],
     // temp during mig
```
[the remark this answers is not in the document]
bodinsamuel: indeed 👌🏻

src/utils/request.ts
bodinsamuel: It's a matter of strictness, in the crawler we got hit by strong bug because typescript is good a guessing the return type and does not warn you if an other function expected a "compatible but slightly different type".

src/watch.ts
bodinsamuel: actually it was wrong, there is no limit as we can only receive one per one

Haroenv (approved):
let's go!

bodinsamuel:
🤞🏻

PR #636, opened 2021-07-02
fix: add new worker to bootstrap

bodinsamuel:


package.json
```diff
@@ -14,7 +14,8 @@
     "doctoc": "doctoc --notitle --maxlevel 3 README.md",
     "lint": "eslint --ext=jsx,ts,tsx,js .",
     "prepare": "husky install",
-    "start": "UV_THREADPOOL_SIZE=12 node -r dotenv/config --async-stack-traces --max-old-space-size=920 dist/index.js",
+    "start": "UV_THREADPOOL_SIZE=12 node -r dotenv/config --async-stack-traces --max-old-space-size=420 dist/index.js",
+    "start_new": "indexName=npm-search-new bootstrapIndexName=npm-search-new.tmp UV_THREADPOOL_SIZE=22 node --async-stack-traces --max-old-space-size=420 dist/index.js",
```
Haroenv, line 18: Are environment variables global in a heroku app?

Haroenv (approved):

package.json
bodinsamuel: only the one defined in the heroku dashboard, that's why I'm overriding them here.
Should work in theory 😬

PR #638, opened 2021-07-02
fix: deprecated facets should be boolean

bodinsamuel:
`deprecated` currently store the raw string, but it's unnecessary for the faceting and could probably slow down the requests/aggregations.

If the field is used for display (I think it is in yarnpkg), you will need to use `deprecatedReason`

Screenshot 2021-07-02 at 15 02 30


src/formatPkg.ts
```diff
@@ -157,7 +159,8 @@ export default function formatPkg(pkg: GetPackage): RawPkg | undefined {
     gitHead: githubRepo ? githubRepo.head : null, // remove this when we update to the new schema frontend
     readme: pkg.readme,
     owner,
-    deprecated: cleaned.deprecated !== undefined ? cleaned.deprecated : false,
+    deprecated: isDeprecated,
+    deprecatedReason: isDeprecated ? String(cleaned.deprecated) : null,
```
Haroenv, line 163: I'm not sure we can afford breaking changes on the index, can you keep `deprecated` as is, and add a new field `isDeprecated` for the boolean form?

bodinsamuel: We could, we are providing this for free and it was not good. 
However we can indeed provide a grace period for the transition, by adding not changing `deprecated` for the moment.
Do we have a list of authorised user somewhere to contact?

bodinsamuel: I have added the new field and put back deprecated

Haroenv (approved):
let's go with this as a start :)

PR #639, opened 2021-07-02
feat(typescript): pre-load definitely typed pkg

bodinsamuel:
Copied a bit from #407
But kept most of the logic around "type check" similar, to avoid to many changes


src/typescript/index.ts
```diff
@@ -0,0 +1,108 @@
...
+    };
+  }
+
+  if (pkg.types.ts === false) {
+    return { types: { ts: false } };
+  }
+
+  // Do we have a main .d.ts file?
+  // TO DO: replace this with a list of files check
+  if (pkg.types.ts !== 'definitely-typed' && pkg.types.ts.possible === true) {
```
Haroenv, line 82: how can ts be definitely-typed here?

src/typescript/index.test.ts
```diff
@@ -0,0 +1,143 @@
...
+        name: 'my-lib',
+        types: { ts: { possible: true, dtsMain: 'main.d.ts' } },
+        version: '1.0.0',
+      });
+      expect(typesSupport).toEqual({ types: { ts: 'included' } });
+    });
+  });
+
+  // TO DO : reup this
+  // adescribe('FilesList', () => {
```
Haroenv, line 107: what's missing to add these tests back?

bodinsamuel: In the aforementioned PR I was prefetching the list of files beforehand but it requires more work and it's not the slowest part of the code

src/typescript/index.ts
bodinsamuel: It can not but Typescript is messing with me, I prefer to have an useless if than adding `// @ts-ignore`

Haroenv (approved):

PR #643, opened 2021-07-04
fix: add expiresAt field

bodinsamuel:
Add field that log when package should be recrawled.
The logic is basic for now:
- 30 days for every packages
- 7 days if it is popular

The goal in a futur PR is to use this facet to easily query outdated package.
Because we can not query with SQL logic like "packages < DATE", the easiest will be to query the facets values, iterate on the values and filter one by one each value.


src/@types/pkg.ts
```diff
@@ -72,6 +72,7 @@ export interface RawPkg {
   types: TsType;
   moduleTypes: ModuleType[];
   lastCrawl: string;
+  expiresAt: string;
```
Haroenv, line 75: we can put it in _searchInternal to avoid exposing it too much?

src/npm/index.ts
```diff
@@ -283,6 +286,11 @@ async function getDownloads(pkgs: Array<Pick<RawPkg, 'name'>>): Promise<
       humanDownloadsLast30Days: numeral(downloadsLast30Days).format('0.[0]a'),
       downloadsRatio,
       popular,
+      expiresAt: popular
+        ? new Date(Date.now() + config.populareExpiresAt)
```
Haroenv, line 290:
```suggestion
        ? new Date(Date.now() + config.popularExpiresAt)
```

src/config.ts
```diff
@@ -164,6 +166,8 @@ export const config = {
   indexSettings,
   indexSynonyms,
   indexRules,
+  expiresAt: 30 * 24 * 3600 * 1000,
+  populareExpiresAt: 7 * 24 * 3600 * 1000,
```
Haroenv, line 170:
```suggestion
  popularExpiresAt: 7 * 24 * 3600 * 1000,
```

Haroenv, line 170: these can be configured with `ms`, for consistency with the rest :)

Haroenv (commented):
great idea!

src/config.ts
bodinsamuel: good point 👌🏻

src/@types/pkg.ts
bodinsamuel: 👌🏻

Haroenv (approved):

src/@types/pkg.ts
```diff
@@ -75,6 +75,7 @@ export interface RawPkg {
   moduleTypes: ModuleType[];
   lastCrawl: string;
   _searchInternal: {
+    expiresAt: string;
```
Haroenv, line 78: if we want to use it for a filter, why isn't it a timestamp/number?

bodinsamuel: same same, it's just easier to read as string no?

PR #646, opened 2021-07-06
feat(requests): add user-agent and httpsAgent

bodinsamuel:
Following @MartinKolarik answers, we were indeed lacking transparency in the user-agent.

Added in:
- All http requests
- Algoliasearch
- Nano

Took the opportunity to pool the agent for:
- algoliasearch
- nano

This should help better controlling the network throughput

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,8 @@
     "doctoc": "doctoc --notitle --maxlevel 3 README.md",
     "lint": "eslint --ext=jsx,ts,tsx,js .",
     "prepare": "husky install",
-    "start": "UV_THREADPOOL_SIZE=12 node -r dotenv/config --async-stack-traces --max-old-space-size=920 dist/index.js",
-    "start_new": "indexName=npm-search-new bootstrapIndexName=npm-search-new.tmp UV_THREADPOOL_SIZE=22 node --async-stack-traces --max-old-space-size=920 dist/index.js",
+    "start": "UV_THREADPOOL_SIZE=64 node -r dotenv/config --async-stack-traces --max-old-space-size=920 dist/index.js",
+    "start_new": "indexName=npm-search-new bootstrapIndexName=npm-search-new.tmp UV_THREADPOOL_SIZE=64 node --async-stack-traces --max-old-space-size=920 dist/index.js",
     "test:watch": "jest --watchAll --no-watchman",
     "test": "jest --forceExit"
   },
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "dependencies": {
     "@algolia/requester-node-http": "4.10.3",
     "@sentry/node": "6.5.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -14,8 +14,8 @@
     "doctoc": "doctoc --notitle --maxlevel 3 README.md",
     "lint": "eslint --ext=jsx,ts,tsx,js .",
     "prepare": "husky install",
-    "start": "UV_THREADPOOL_SIZE=12 node -r dotenv/config --async-stack-traces --max-old-space-size=920 dist/index.js",
-    "start_new": "indexName=npm-search-new bootstrapIndexName=npm-search-new.tmp UV_THREADPOOL_SIZE=22 node --async-stack-traces --max-old-space-size=920 dist/index.js",
+    "start": "UV_THREADPOOL_SIZE=64 node -r dotenv/config --async-stack-traces --max-old-space-size=920 dist/index.js",
+    "start_new": "indexName=npm-search-new bootstrapIndexName=npm-search-new.tmp UV_THREADPOOL_SIZE=64 node --async-stack-traces --max-old-space-size=920 dist/index.js",
```
Haroenv, line 18: what's the thread pool thing?

src/utils/request.ts
```diff
@@ -4,6 +4,11 @@ import https from 'https';
 import type { OptionsOfJSONResponseBody } from 'got';
 import got from 'got';
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires, import/no-commonjs
+const { version } = require('../../package.json');
```
Haroenv, line 8: do we actually update the package.json?

bodinsamuel: not right now, but we could !

package.json
bodinsamuel: Sorry it was not in the PR desc but in the commit b44246e
`increase UV to help DNS resolution` LIB_UV is responsible for DNS resolution (and others things) by default it's 4 so a bit short for heavy HTTP script

package.json
```diff
@@ -27,6 +27,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@algolia/requester-node-http": "4.10.3",
     "@sentry/node": "6.5.1",
     "algoliasearch": "4.9.3",
```
Haroenv, line 32: synchronise the requester and the client

Haroenv (commented):
👌

package.json
bodinsamuel: expose everything in algoliasearch 😬

Haroenv: https://github.com/algolia/algoliasearch-client-javascript/pull/1284 among others I guess, but I agree, should be exposed :)

bodinsamuel: <3 This will help us a lot on the crawler

PR #647, opened 2021-07-07
fix: improve logging + remove catchup

bodinsamuel:
- Catchup was wrong, changes can not be processed in batch.
I have changed the behavior during the typescript migration but now I can delete it and just leave the watch.

- Add better logging to understand what is going on


Haroenv (approved):
looks like this will be easier, although I'm not sure what the actual issue is with replicate

bodinsamuel:
> the actual issue is with replicate

Currently nothing, before yes. 
It was batching changes, but in a single batch there could be the same package, like multiple update or a create/delete.
Processing them in the wrong order could lead to corrupted data, so in the end the safest way was to process 1:1 which is the same as running the watch.

Haroenv:
makes perfect sense!

PR #648, opened 2021-07-07
feat: prepare docker

bodinsamuel:
- Add semantic release to increment version
- Add Dockerfile that compiles the code and only keep what is needed in the final image
- Add automatic release and automatic build in CI (will be tested on master)

Will push image here:
https://hub.docker.com/repository/docker/algolia/npm-search/tags?page=1&ordering=last_updated

## Test

```sh
# Do not use the prod app id :D
docker run -e appId="appid -e apiKey="apikey" algolia/npm-search
```


.circleci/config.yml
```diff
@@ -28,10 +40,32 @@ jobs:
...
+  # Step release
+  release:
+    docker:
+      - image: cimg/node:14.16.1
+    working_directory: ~/repo
+    steps:
+      - attach_workspace: *attach_work_dir
+      - run:
+          name: Release
+          command: yarn semantic-release
```
Haroenv, line 56: is it possible that this will change the version number to something else than what's committed?

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# Install dev dependencies
+RUN true \
+  # && yarn set version berry \
+  && yarn install --production=false
+
+# This step will invalidates cache
+COPY . /app
+RUN ls -lah /app
+
+# Builds the UI install only the prod dependencies
```
Haroenv, line 22: what's "the UI"?

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# Do not use root to run the app
+USER node
+
+WORKDIR /app
+
+COPY --from=base --chown=node:node /app /app
+
+EXPOSE 8000
+
+CMD [ "npm", "start" ]
```
Haroenv, line 43: I think currently we run yarn start, and there's the start_new command too, can this be configurable from docker or is it not needed?

package.json
```diff
@@ -10,14 +10,16 @@
...
     "lint": "eslint --ext=jsx,ts,tsx,js .",
     "prepare": "husky install",
-    "start": "UV_THREADPOOL_SIZE=64 node -r dotenv/config --async-stack-traces --max-old-space-size=920 dist/index.js",
+    "start": "UV_THREADPOOL_SIZE=64 node --async-stack-traces --max-old-space-size=920 dist/index.js",
     "start_new": "indexName=npm-search-new bootstrapIndexName=npm-search-new.tmp UV_THREADPOOL_SIZE=64 node --async-stack-traces --max-old-space-size=920 dist/index.js",
     "test:watch": "jest --watchAll --no-watchman",
-    "test": "jest --forceExit"
+    "test": "jest --forceExit",
+    "docker:build": "./scripts/build.sh",
+    "docker:release": "VERSION=$(npx json -f package.json version) && docker push algolia/npm-search && docker push algolia/npm-search:$VERSION"
```
Haroenv, line 22: don't we want json to be a dependency if used in a script?

Haroenv (commented):
I don't have docker on my machine, so didn't try it out yet, will this be used on heroku yet or not?

bodinsamuel:
> I don't have docker on my machine, so didn't try it out yet, will this be used on heroku yet or not?

Nope, it's will be used for testing on GCP (and make it available for anyone in one click too)

.circleci/config.yml
bodinsamuel: I'm not sure to get your question?

Dockerfile
bodinsamuel: bad c/c 😬

Dockerfile
bodinsamuel: we currently launch a different command to pass different env vars in the same heroku instance but it won't be needed in GCP as you just need to pass env var in the command line.

package.json
bodinsamuel: it could, we usually don't do it because it's not a requirement to run and dev on the project

.circleci/config.yml
Haroenv: I misunderstood what semantic-release does

package.json
Haroenv:
```suggestion
    "docker:release": "VERSION=$(node -e 'console.log(require(\'./package.json\').version)') && docker push algolia/npm-search && docker push algolia/npm-search:$VERSION"
```

What do you think of this version instead? That way nothing can go wrong with the dependency

bodinsamuel: allez

Haroenv (commented):
lgtm!

PR #650, opened 2021-07-10
feat: add health API

bodinsamuel:
Add health API to be able to load the script in monitored system

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,0 +1,18 @@
+import http from 'http';
+
+import { log } from './utils/log';
+
+export function createAPI(): void {
+  const server = http.createServer((req, res) => {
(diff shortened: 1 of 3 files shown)


src/api.ts
```diff
@@ -0,0 +1,18 @@
+import http from 'http';
+
+import { log } from './utils/log';
+
+export function createAPI(): void {
+  const server = http.createServer((req, res) => {
```
Haroenv, line 6: do we know what the overhead of this server will be?

bodinsamuel: there is not much overhead for a simple open connection,
It's not exposed to the public it's just for internal health check, pinged every 10-30s

PR #653, opened 2021-07-11
fix: cache total downloads

bodinsamuel:
It's not the longest query, but it's still not useful to do it at every batch.


Haroenv (commented):
good idea, it could even be cached for longer I think

PR #656, opened 2021-07-11
feat: process package in queue instead of batch

bodinsamuel:
Most batch were slowed down by the slowest package, which means fast packages were still not indexed.
Here, we take advantage of NodeJs being good with async/sync and indexing sooner.

The prefetcher, only prefetch the names of the packages, so it's not filling the memory but still provide a way to fill the queue rapidly.

In the queue we fetch package per package, which is a bit slower than fetching a pages of doc, but results in less memory used and gc sooner too.
The processing itself is the same as before, except we don't do batch but one by one (see #652) 

--- 

Except that, I have been extensively testing in Google Compute Engine, to replace slow and expensive Heroku.
I lack a deploy script, monitoring and everything else, but it's working very fine (see comments below). 
GCE being very "manual" (except the start of the docker image and the CPU tracking there is nothing 🤦🏻 ), I think I'll go for the overkill solution -> Kubernetes, but at least it has native monitoring, auto-restart and we know the stack perfectly at Algolia.

### DIFF
--- a/src/__tests__/saveDocs.test.ts
+++ b/src/__tests__/saveDocs.test.ts
@@ -213,7 +213,7 @@ it('should be similar batch vs one', async () => {
 
   const row = { id: '', key: 'preact', value: { rev: 'a' }, doc: preact };
   await saveDocs({ docs: [row], index });
-  await saveDoc({ row, index });
+  await saveDoc({ row: preact, index });
 
   expect(index.saveObjects).toHaveBeenCalledWith([clean]);
   expect(index.saveObject).toHaveBeenCalledWith(clean);
--- a/src/bootstrap.ts
+++ b/src/bootstrap.ts
@@ -170,5 +170,5 @@ function createPkgConsumer(
     } finally {
       datadog.timing('loop', Date.now() - start);
     }
-  }, 3);
+  }, config.bootstrapConcurrency);
 }
--- a/src/bootstrap.ts
+++ b/src/bootstrap.ts
@@ -1,13 +1,14 @@
 import type { SearchClient, SearchIndex } from 'algoliasearch';
 import type { QueueObject } from 'async';
 import { queue } from 'async';
+import chalk from 'chalk';
 
 import type { StateManager } from './StateManager';
 import * as algolia from './algolia';
 import { config } from './config';
 import * as npm from './npm';
+import type { PrefetchedPkg } from './npm/Prefetcher';
 import { Prefetcher } from './npm/Prefetcher';
-import type { GetPackage } from './npm/types';
 import { saveDoc } from './saveDocs';
 import { datadog } from './utils/datadog';
 import { log } from './utils/log';
@@ -15,6 +16,7 @@ import * as sentry from './utils/sentry';
 import { wait } from './utils/wait';
 
 let prefetcher: Prefetcher;
+let consumer: QueueObject<PrefetchedPkg>;
 
 /**
  * Bootstrap is the mode that goes from 0 to all the packages in NPM
@@ -28,7 +30,7 @@ let prefetcher: Prefetcher;
  * Watch mode should/can be reliably left running for weeks/months as CouchDB is made for that.
  * BUT for the moment it's mandatory to relaunch it because it's the only way to update: typescript, downloads stats.
  */
-async function run(
+export async function run(
   stateManager: StateManager,
   algoliaClient: SearchClient,
   mainIndex: SearchIndex,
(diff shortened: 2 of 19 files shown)


bodinsamuel:
Currently processing in GCP

bodinsamuel:
Latest number are good, we did x6 increase, **from ~1 pkg/s to ~6 pkg/s** 
At this rate we have saved a full week and will be done by the end of the day.
Screenshot 2021-07-12 at 11 01 37

There are still bottlenecks, probably due to the nature of some packages, for example here at 6am everything cache missed in jsdelivr.
Screenshot 2021-07-12 at 11 04 23

---

In term of processing, I'm cheating a bit with a preemptible e2-medium, but for 2days of intensive usage we are paying $1.35 which is far better than heroku. About 40% less, not even considering I have x4 times the computational power. 
The same configuration would probably cost around 300$ per month vs 40-50$ now; we'll see by the end of the month.
n.b: the most expensive right now is not the CPU but the network.

src/__tests__/saveDocs.test.ts
```diff
@@ -213,7 +213,7 @@ it('should be similar batch vs one', async () => {
 
   const row = { id: '', key: 'preact', value: { rev: 'a' }, doc: preact };
```
Haroenv, line 214: is this variable still used?

src/npm/Prefetcher.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      return this.#ready.shift()!;
+    }
+  }
+
+  async launch(): Promise<void> {
+    this.#running = true;
+    while (this.#running) {
+      if (this.#ready.length >= this.#maxIdle) {
+        await wait(5000);
```
Haroenv, line 52: should these be constants or configuration?

src/npm/Prefetcher.ts
```diff
@@ -0,0 +1,84 @@
...
+  async getNext(): Promise<PrefetchedPkg> {
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
+      if (this.#ready.length <= 0) {
+        await wait(100);
+        continue;
+      }
+
+      return this.#ready.shift()!;
+    }
```
Haroenv, line 45:
```suggestion
    while (this.#ready.length <= 0) {
      await wait(100);
    }
    return this.#ready.shift()!;
```

should this not be the condition of the while, and the ready shift after it?

src/bootstrap.ts
```diff
@@ -24,17 +31,20 @@ let loopStart: number = Date.now();
...
  */
-async function run(
+export async function run(
   stateManager: StateManager,
   algoliaClient: SearchClient,
   mainIndex: SearchIndex,
   bootstrapIndex: SearchIndex
 ): Promise<void> {
+  log.info('-----');
+  log.info('⛷   Bootstrap: starting');
```
Haroenv, line 41: in this PR you remove some of the log emojis (like in run), should they all be removed as it's not consistent anymore, added back, or is it fine with a mix?

src/bootstrap.ts
```diff
@@ -54,70 +64,53 @@ async function run(
...
-
-    lastProcessedId = await loop(lastProcessedId, stateManager, bootstrapIndex);
-  } while (lastProcessedId !== null);
-
-  log.info('-----');
-  log.info('⛷   Bootstrap: done');
-  await stateManager.save({
-    bootstrapDone: true,
-    bootstrapLastDone: Date.now(),
+  prefetcher = new Prefetcher({
```
Haroenv, line 70: this utility sounds like it could be something built into async already, do you know if it exists?

Haroenv (commented):
looks like a solid refactor, a couple detail questions though

src/__tests__/saveDocs.test.ts
bodinsamuel: yes the line just after in saveDocs.
saveDocs will be removed after I deploy

src/npm/Prefetcher.ts
bodinsamuel: good point, much cleaner 👌🏻

src/bootstrap.ts
bodinsamuel: I removed some to focus on readability, just leaving on the main logs. I'm good with a mix, it highlight big steps

src/bootstrap.ts
bodinsamuel: independently they are some functions that could work like `whilst` + `cargo` or `queue` + `until` 
I don't think there is one function that execute until a max and continue when an other condition occurs.

If I use `until` I still the code around that limits the prefetch, wait between iteration, save the offset and the fetch itself. 
I might save 10lines top

Haroenv (approved):
let's go with this implementation, exciting!

PR #658, opened 2021-07-12
fix: stop using unpkg for changelog

bodinsamuel:
Now that we pass jsdelivr file list there is no point in loading the same data less efficiently from unpkg.

### DIFF
--- a/src/__tests__/changelog.test.ts
+++ b/src/__tests__/changelog.test.ts
@@ -6,7 +6,6 @@ jest.mock('got', () => {
     'https://raw.githubusercontent.com/algolia/algoliasearch-netlify/master/CHANGELOG.md',
     'https://bitbucket.org/atlassian/aui/raw/master/changelog.md',
     'https://raw.githubusercontent.com/expressjs/body-parser/master/HISTORY.md',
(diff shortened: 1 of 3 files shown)


src/__tests__/changelog.test.ts
```diff
@@ -6,7 +6,6 @@ jest.mock('got', () => {
     'https://raw.githubusercontent.com/algolia/algoliasearch-netlify/master/CHANGELOG.md',
     'https://bitbucket.org/atlassian/aui/raw/master/changelog.md',
     'https://raw.githubusercontent.com/expressjs/body-parser/master/HISTORY.md',
-    'https://unpkg.com/@atlaskit/button@13.3.7/CHANGELOG.md',
```
Haroenv, line 9: IRL this one comes from jsdelivr now right?

Haroenv (commented):
looks solid, this code indeed is replaced with the /flat part of jsdelivr

src/__tests__/changelog.test.ts
Haroenv: I see, it's available in jsDelivr so not a regression

Haroenv (approved):

PR #663, opened 2021-07-13
fix: various

bodinsamuel:
- Increase bootstrap period because it should be less needed now and it takes too much time to reprocess that often
- Log progress to datadog to remove the need to read logs to see progress
- Log percent in 00.00% format for more precision
- Drain queue before moving index and await the copy (we missed the last few packages in the bootstrap)

### DIFF
--- a/src/bootstrap.ts
+++ b/src/bootstrap.ts
@@ -96,7 +96,7 @@ export async function run(
 
   consumer.pause();
 
-  // The while will stop before the queue has been fully consumed
+  // While we no longer are in "processing" mode, it's possible that there's a last iteration in the queue
   await consumer.drain();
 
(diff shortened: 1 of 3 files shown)


src/bootstrap.ts
```diff
@@ -96,15 +96,19 @@ export async function run(
 
   consumer.pause();
 
-  log.info('-----');
-  log.info('⛷   Bootstrap: done');
-  log.info('-----');
+  // The while will stop before the queue has been fully consumed
```
Haroenv, line 99:
```suggestion
  // While we no longer are in "processing" mode, it's possible that there's a last iteration in the queue
```

Haroenv (approved):
good fix

PR #664, opened 2021-07-14
fix: expiresAt can be a numericFilter

bodinsamuel:
When doing #643 I was thinking we can not do "Date < OtherDate" but it's actually doable with number 🤦🏻 
Changing date to number, removing the facet as a simple numeric filter can be used here.


Haroenv (approved):
that's what I said in the PR 😢 https://github.com/algolia/npm-search/pull/643#discussion_r663840181

bodinsamuel:
> that's what I said in the PR 😢 https://github.com/algolia/npm-search/pull/643#discussion_r663840181

Ahah yes but I understood it as "replacing the string by a number but still using facets"

PR #668, opened 2021-07-19
feat: upgrade deps

bodinsamuel:
manual upgrade because we are late on a lot of things not sure why.
and I needed a good reason to increase the version to a new major


package.json
```diff
@@ -59,36 +59,36 @@
...
     "eslint-plugin-eslint-comments": "3.2.0",
     "eslint-plugin-import": "2.23.4",
     "eslint-plugin-jasmine": "4.1.2",
     "eslint-plugin-jest": "24.3.6",
-    "eslint-plugin-jsdoc": "35.4.1",
+    "eslint-plugin-jsdoc": "35.4.5",
     "eslint-plugin-prettier": "3.4.0",
     "eslint-plugin-react": "7.24.0",
-    "husky": "6.0.0",
+    "husky": "7.0.1",
```
Haroenv, line 82: isn't this the one with the wild license?

bodinsamuel: not sure what you are talking about, husky is MIT

Haroenv: https://github.com/typicode/husky/issues/857 it was only v5

PR #683, opened 2021-08-18
feat: add package style types information

zarianec:
Hey there. We are using Algolia as a search engine on the jsDelivr website and we have some ideas on how to make it even better. With this PR I would like to initiate the discussion and provide the starting point for possible implementation.

So, what we are actually proposing:

## Introduce the `styleTypes` field

The index currently contains a `moduleTypes` field, which is an array of multiple values containing all detected module types (esm, cjs, unknown at the moment). The goal here is to add a similar field for filtering packages that have CSS.

The primary focus is on plain CSS files, but to make this more useful, we may as well detect other types like less/sass. I.e., `styleTypes: [ 'css', 'less' ]` would indicate that there are css and less files in the package.

The proposal for `styleTypes` field:
- css - css exist based on listing,
- less - less files exist based on listing,
- sass - sass files exist based on listing,
- any other style files?
- none - no styles in this package - it's purely JS

The `styleTypes` should be detected from both package.json's `style` field and package's file listing.

## Make `moduleTypes` more flexible

The current combination of values for `moduleTypes` doesn't allow making one very useful type of query: "all packages, that contain JS code" - that currently classifies as unknown, and even adding the `styleTypes` field won't help because styles are not mutually exclusive with js.

We propose updating this field with a new value of `none`, so together it is:
- esm - same meaning as now, based on package info,
- cjs - same meaning as now, based on package info,
- none - there are no JS files in the package, i.e., we're 100% sure this isn't a JS package,
- unknown - the rest - there are some JS files but we don't know the type exactly.

`none` should be based on file listing and consider all typically used js extensions, i.e, js, mjs, cjs (any other?).

## Why

A combination of `moduleTypes` and `styleTypes` will allow making more precise filtering not available right now.

Specifically, it'll be possible to make all of these queries:
- `moduleTypes: esm` - packages that are certainly esm,
- `NOT moduleTypes: none` - packages that contain any kind of js code,
- `styleTypes: css` - packages that have plain css,
- `NOT styleTypes: none` - packages that have any kind of css.

Particularly the `NOT moduleTypes: none` option will be useful for most JS searches.

---

### Pushing it further

Feature, described below is not a part of this PR but information to think about.

We just released a new feature for jsDelivr API, that allows you to resolve the entry points for a package. It supports both js and css and based on package.json file as well as advanced heuristics for packages where entry points not explicitly set in package.json.
Having main files for a package in the index probably may be useful as well.

Example of resolved entries:
```
bootstrap@5.1.0:
{
	"js": {
		"file": "/dist/js/bootstrap.min.js",
		"guessed": false
	},
	"css": {
		"file": "/dist/css/bootstrap.min.css",
		"guessed": false
	}
}
```


MartinKolarik:
Note that we decided to use the file listings here because many packages do not use the `style` field -  the current approach is similar to what we used in the new API endpoint mentioned above, and we found it to work fairly well.

src/pkgTypes.ts
```diff
@@ -0,0 +1,94 @@
...
+  pkg: Pick<RawPkg, 'styleTypes'>,
+  filelist: File[]
+): Pick<RawPkg, 'styleTypes'> {
+  const start = Date.now();
+
+  try {
+    const styleTypes = new Set(pkg.styleTypes);
+
+    for (const file of filelist) {
+      if (!createFilePattern(styleFileExtensions).test(file.name)) {
```
MartinKolarik, line 25: Why are we constructing a new regexp here for each file?

src/formatPkg.ts
```diff
@@ -634,3 +638,17 @@ function getModuleTypes(pkg: NicePackageType): ModuleType[] {
...
 }
+
+function getStyleTypes(pkg: NicePackageType): StyleType[] {
+  // style not declared - we will detect it later based on file list
+  if (!pkg.style) {
+    return [];
+  }
+
+  const ext = pkg.style.split('.').pop() as StyleType;
+  if (ext && styleFileExtensions.includes(ext)) {
```
Haroenv, line 649: maybe instead of detecting specific extensions, we should save _all_ extensions?

src/pkgTypes.ts
zarianec: my bad, I'll move it outside the loop

src/saveDocs.ts
```diff
@@ -97,9 +103,11 @@ async function addMetaDatas(pkgs: RawPkg[]): Promise<FinalPkg[]> {
     jsDelivr.getAllFilesList(pkgs),
   ]);
 
-  const [changelogs, ts] = await Promise.all([
+  const [changelogs, ts, moduleTypes, styleTypes] = await Promise.all([
     getChangelogs(pkgs, filelists),
     getTSSupport(pkgs, filelists),
+    getModuleTypesForAll(pkgs, filelists),
```
Haroenv, line 109: this now calculates moduleTypes both during formatPkg and here, won't one always override the other?

src/formatPkg.ts
zarianec: the idea was to filter out possible "trash" in the `style` field if the author set non-style file there. Not sure if it's necessary but it's just a preventive action here.

MartinKolarik: That's an interesting idea but without the list of extensions of specific types, we wouldn't be able to set the `none` value - which I think provides some convenience, searching for `NOT none` vs `js OR cjs OR esm` - but I suppose you could also argue that it's better the user specifies the extensions they want explicitly.

src/saveDocs.ts
zarianec: If `moduleTypes` is resolved to anything different from `unknown` in formatPkg (package.json based), no further calculation will be made. For this field, it's enough to know that any of `esm` or `cjs` is detected.
But if it's `unknown` we must check if there are js files at all.

zarianec: But this will not cover the case when package.json is:

src/formatPkg.ts
MartinKolarik: I can see some additional possibilities it would bring, e.g. searching for packages with web fonts, but there's again the question, if we want to make it low level and let users specify all extensions, or if we give them prepared categories (which could be done for fonts too).

Haroenv: you can deal with none by defaulting to `styleType: ['none']`, as that can be filtered for, although then you're overlapping between possibly genuine extension called "none" (likely a non-issue)

src/saveDocs.ts
Haroenv: invalid packages like that is a complicated situation. I think it's probably okay to consider that an anomaly and leave it as "undefined behaviour". In this case that would be something like marking it as cjs when really we don't know what type it is is probably fine

src/pkgTypes.ts
zarianec: https://github.com/algolia/npm-search/pull/683/commits/7df273f9c2b1292db4920092b3693102efb2a5d4

src/formatPkg.ts
MartinKolarik: Oh, I misunderstood you meant just for the `style` field (thought you wanted a more generic version that would have all extensions of all files in the package) - since we combine the information from `style` with the information based on file listings, I think it makes sense to have a static set of expected values. If we find there's a new value that makes sense, it should be added to the file listing detection as well.

Haroenv: how would we find out which other extensions could make sense? maybe `.styl` has some usage, or something new would come up. If we base it on the data instead of a static list, we won't miss any extensions in the future if they become popular. What do you think?

MartinKolarik: Ok, that's a fair point - let's allow all extensions for the `style` field, just normalize it to lowercase.

zarianec: fixed here https://github.com/algolia/npm-search/pull/683/commits/ab397697220d928bfce2b8013ddc54c17f1d6592

jimaek:
@Haroenv Is there anything we need to do to get the CI tests running?

Haroenv:
I think they won't run for external contributors, I'd need to check if there's some setting that needs to be changed @jimaek, this repo hasn't gotten any external contributions since we switched from travis to circle recently I think

src/npm/types.ts
```diff
@@ -45,6 +45,7 @@ export interface GetVersion {
   schematics?: string;
   types?: string;
   typings?: string;
+  style?: string;
```
bodinsamuel, line 48: The type is currently in bad state to comply with old code during migration (I should add a warning). 
Style/typings and such are not returned in the root level but in the `versions` key.
See Bootstrap for example: https://registry.npmjs.com/bootstrap

NicePackage support merging root level / versions key, but it's incorrect; this will change a bit your tests but not the logic.

src/pkgTypes.ts
```diff
@@ -0,0 +1,97 @@
+import type { RawPkg, StyleType } from './@types/pkg';
```
bodinsamuel, line 1: Can it be moved to `jsDelivr/`? as it is strictly depend on the fileList provided by the service

src/__tests__/pkgTypes.test.ts
```diff
@@ -0,0 +1,97 @@
...
+          [{ name: '/dist/style/style.min.css' }],
+          [
+            { name: '/src/style/style.less' },
+            { name: '/dist/style/style.min.css' },
+            { name: '/dist/js/lib.min.js' },
+            { name: '/style.scss' },
+          ],
+          undefined,
+          [{ name: '/src/style/style.less' }],
+        ] as File[][]
```
bodinsamuel, line 26: Usually not a fan of using `as Type`, this is handy for testing but usually makes the job of refactoring harder because TS will not complain if something is missing. 
What I favor is using a base object that I can spread and reuse, it decrease a bit the readability but will help future me/you.

src/formatPkg.ts
```diff
@@ -634,3 +637,14 @@ function getModuleTypes(pkg: NicePackageType): ModuleType[] {
 
   return [...moduleTypes];
 }
+
+function getStyleTypes(pkg: NicePackageType): StyleType[] {
```
bodinsamuel, line 641: Is this relevant to do it at this stage if you always override the information later?
Especially if values is not filtered (cf: .woff)

src/pkgTypes.ts
```diff
@@ -0,0 +1,97 @@
...
+import type { File } from './jsDelivr';
+import { datadog } from './utils/datadog';
+
+const styleFileExtensions: StyleType[] = ['css', 'less', 'scss'];
+const jsFileExtensions = ['js', 'mjs', 'cjs'];
+
+function createFilePattern(extensions: string[]): RegExp {
+  const extPattern = extensions.join('|');
+  return new RegExp(
+    `^(?:(?!\\/(docs?|documentation|examples?|samples?|demos?|tests?)\\/)(?!\\/[._]).)+\\.(${extPattern})$`
```
bodinsamuel, line 11: Can be nice to provide a regex101 link to see what you are looking to exclude/include in this; negative look ahead/behind are hard to understand :p

src/pkgTypes.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export function getStyleTypes(
+  pkg: Pick<RawPkg, 'styleTypes'>,
+  filelist: File[]
+): Pick<RawPkg, 'styleTypes'> {
+  const start = Date.now();
+
+  try {
+    const styleTypes = new Set(pkg.styleTypes);
+    const filePattern = createFilePattern(styleFileExtensions);
```
bodinsamuel, line 23: I don't you need to build the file pattern at each execution, just a constant is enough no?

bodinsamuel (commented):
Hey,

Thanks for this PR, this looks very good 🙏🏻 I'm glad we did the TS migration and the file list feature, add much more confidence when reviewing.

I have a few comments mostly around the form, but I feel like the idea is great, very straightforward.
Checked the CI for external Pull Request, it should run now.

src/npm/types.ts
zarianec: I'm not sure I understood you correctly so let me clarify this - it's fine to have the `style` field in the GetVersion interface. But I violate that interface in my tests and that's not good 🙂 

I must change the tests to follow the interface:
```typescript
expect(
  formatPkg({
    ...BASE,
    versions: { // this field with its subfields was missing in original tests
      '1.0.0': { ...BASE_VERSION, style: '/style.min.css' }
    },
  })?.styleTypes
).toEqual(['css']);
```

src/@types/pkg.ts
```diff
@@ -32,7 +32,9 @@ export type TsType =
       definitelyTyped: string;
     };
 
-export type ModuleType = 'esm' | 'cjs' | 'unknown';
+export type ModuleType = 'esm' | 'cjs' | 'none' | 'unknown';
+
+export type StyleType = 'css' | 'less' | 'scss' | 'none';
```
Haroenv, line 37: has this become `string` now?

src/__tests__/pkgTypes.test.ts
Haroenv: agreed, usually something like `createFile(partial)` should deal with the default values needed

src/formatPkg.ts
Haroenv: Why does it get overridden? isn't the other value packageType?

src/pkgTypes.ts
```diff
@@ -0,0 +1,97 @@
+import type { RawPkg, StyleType } from './@types/pkg';
+import type { File } from './jsDelivr';
+import { datadog } from './utils/datadog';
+
+const styleFileExtensions: StyleType[] = ['css', 'less', 'scss'];
+const jsFileExtensions = ['js', 'mjs', 'cjs'];
```
Haroenv, line 6: maybe what we should be looking for is all extensions that aren't in test, documentation or examples. That can then be used frontend for a filter without being opinionated about what's a script and what's a style

src/npm/types.ts
bodinsamuel: yes exactly (sorry I was not really clear)
style only exists in the GetVersion interface

src/formatPkg.ts
zarianec: @bodinsamuel It doesn't get overridden but updated with new values. 
Here I get the file extension from the `style` field of package.json and letter add the extensions of style files from a listing.

src/@types/pkg.ts
zarianec: https://github.com/algolia/npm-search/pull/683/commits/80bb5089760c3fbde5e182da067ad8a9c1f49a78

src/pkgTypes.ts
zarianec: https://github.com/algolia/npm-search/pull/683/commits/8ff01012a86471b8ba3a013efd7c8731f1557ee8

src/npm/types.ts
zarianec: https://github.com/algolia/npm-search/pull/683/commits/4be334b8d4fd2ae7ddfb3728b9bdcf7a5cd7dc90

src/pkgTypes.ts
zarianec: fixed in https://github.com/algolia/npm-search/commit/8ff01012a86471b8ba3a013efd7c8731f1557ee8

zarianec:
@Haroenv @bodinsamuel ping. Is there anything I could do to move forward with this PR?

bodinsamuel (approved):
Sorry I had missed the notification. 
Looks very good to me 🙌🏻

src/jsDelivr/pkgTypes.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    for (const file of filelist) {
+      if (!styleFilePattern.test(file.name)) {
+        continue;
+      }
+
+      const type = file.name.split('.').pop();
+
+      if (type) {
+        styleTypes.add(type);
```
MartinKolarik, line 38: This should be lowercase as well.

MartinKolarik: Actually, it is now, because the regex isn't case-insensitive - but it should be case-insensitive and normalized to lc.

MartinKolarik:
The schema in README should be updated too.

src/jsDelivr/pkgTypes.ts
zarianec: fixed https://github.com/algolia/npm-search/pull/683/commits/8dd49c3c9ebb44191d4a82cf699bb21d4d86a0d0

README.md
```diff
@@ -134,6 +134,7 @@ For every single NPM package, we create a record in the Algolia index. The resul
     definitelyTyped: '@types/babel__core',
   },
   moduleTypes: ['unknown'], // esm | cjs | unknown
```
MartinKolarik, line 136: esm | cjs | none | unknown

README.md
```diff
@@ -134,6 +134,7 @@ For every single NPM package, we create a record in the Algolia index. The resul
...
+  styleTypes: ['none'], // file extensions like css, less, scss or none if no style files present
```
MartinKolarik, line 137: css | less | scss | none

zarianec: Technically any extension could be there as far as we do not filter files in `style` field

MartinKolarik: Ok, fair.

MartinKolarik (approved):

Haroenv:
going to give this a last review tomorrow

Haroenv (approved):
ok this looks great, sorry for the slow review!

MartinKolarik:
Thanks! How feasible is it to do a reindex now?

PR #781, opened 2021-11-03
fix: put processes in class for state

bodinsamuel:


Haroenv:
Big change :D what's the pros/cons?

bodinsamuel:
Having more and more "floating" state was bothering me + passing all params for each function starts to be too verbose.
Mostly a coding style reason, I'm fine with functional with it's very simple and portable function, here they are all tied together.

bodinsamuel:
Note: it's easier to review with `Hide whitespace` checked

src/watch.ts
```diff
@@ -12,189 +13,196 @@ import { datadog } from './utils/datadog';
...
+   *
+   * @param stateManager - The state manager.
+   * @param mainIndex - Algolia index manager.
+   */
+  async watch(): Promise<boolean | undefined> {
+    const { seq } = await this.stateManager.get();
+    const changesConsumer = this.changesConsumer;
+    if (!changesConsumer) {
+      log.warn('Can not watch without a consumer');
       return;
```
Haroenv, line 85: if we return `undefined`, is the eslint rule disabling still needed?

```suggestion
      return undefined;
```

Haroenv (approved):
code itself lgtm! would like to see if there's a way where the eslint rule doesn't need to be disabled

src/watch.ts
bodinsamuel: There is an issue for that. we could disable the rule entirely for TS though 
https://github.com/typescript-eslint/typescript-eslint/issues/1277

Issue #778, opened 2021-11-02
Valid package missing from the index


MartinKolarik:
[validated-changeset](https://www.npmjs.com/package/validated-changeset) seems to be [completely missing from the index](https://www.jsdelivr.com/package/npm/validated-changeset) (see failed requests) despite existing for quite a while and not being deprecated 🤔

bodinsamuel:
NB: I did not know the Algolia was in the critical path of jsdelivr, maybe we should produce a different index without 3rd party enrichment to avoid disruption.

MartinKolarik:
It doesn't affect our CDN but affects the website since that relies mostly on the data from the index to render the package pages (I think similarly to https://yarnpkg.com/). A retry strategy would definitely be good.

> maybe we should produce a different index without 3rd party enrichment to avoid disruption.

So it was something else than npm that failed?

bodinsamuel:
> So it was something else than npm that failed?

In that case no, NPM was the issue. But 3rd party is what make the indexing slow, you can have many hours of lag because of this. I don't know if it's acceptable for you, considering that without 3rd party the page would probably empty. 

We can eventually think of "on-demand" endpoint, that you can reach to request a package that seems to 404. But I guess you got a lot of 404 that would also spam us. wdyt?

MartinKolarik:
Right now, we're only using the following attributes, which I think all come from npm?

```
[ 'deprecated', 'description', 'githubRepo', 'homepage', 'keywords', 'license', 'name', 'owner', 'version' ]
```

But that's going to change in a redesign that we're working on right now as we also want to show part of READMEs.

We also query our own API which could be used to prevent 404 calls if there was an on-demand endpoint in the future, but the thing is, most people get to the package pages from the search results. If the search doesn't return the package, the on-demand approach won't help. So it would have to be a combination of npm-only index + the endpoint.

I think a good enough start here is ensuring that we don't miss any packages completely because of temporary errors.

Haroenv:
It's not ideal, but a temporary workaround would be in your frontend to fall back to https://registry.npmjs.com/validated-changeset, as that's a more "pure" source of truth

MartinKolarik:
> It's not ideal, but a temporary workaround would be in your frontend to fall back to https://registry.npmjs.com/validated-changeset, as that's a more "pure" source of truth

Indeed for direct access to that page, we could.

bodinsamuel:
Made a quick and dirty retry strategy @MartinKolarik

PR #782, opened 2021-11-04
fix: naive retry strategy

bodinsamuel:
This should help #778 

Two retry strategy:
- Exponential backoff that blocks the queue
  Useful if a service is down, skipped the job right away would just make every other fail faster.
  
- "Mid-term" retry.
  After 3 retries, the jobs are put aside and tried later until they pass or a new update for this package comes in

### DIFF
--- a/src/watch.ts
+++ b/src/watch.ts
@@ -20,7 +20,7 @@ type ChangeJob = { change: DatabaseChangesResultItem; retry: number };
 export class Watch {
   stateManager: StateManager;
   mainIndex: SearchIndex;
-  skipped: ChangeJob[] = [];
+  skipped = new Map<string, ChangeJob>();
   // Cached npmInfo.seq
   totalSequence: number = 0;
   changesConsumer: QueueObject<ChangeJob> | undefined;
@@ -52,6 +52,14 @@ export class Watch {
    *      until an other package is updated.
    *      It will never be up to date because he receive event at the same pace
    *      as they arrive in listener A, even if it's not the same package.
+   *
+   * --- Retry strategy.
+   *  Each packages update needs to be processed in order so there is a first naive strategy that will block the queue.
+   *  It will retry the same package after an exponential backoff, N times.
+   *
+   *  After N times, this update will be discarded in the this.skipped
+   *  This Map will be regularly reprocessed to avoid loosing jobs.
+   *  This is an in-memory only retry, if the process is stopped, skipped job are lost.
    */
   async run(): Promise<void> {
     log.info('-----');
@@ -125,13 +133,16 @@ export class Watch {
   }
 
   /**
-   * Regurarly try to reprocess skipped packages.
+   * Regularly try to reprocess skipped packages.
    */
   checkSkipped(): void {
-    log.info('Checking skipped jobs (', this.skipped.length, ')');
-    datadog.increment('packages.skipped', this.skipped.length);
-    if (this.skipped.length > 0) {
-      const clone = [...this.skipped];
+    log.info('Checking skipped jobs (', this.skipped.size, ')');
+    datadog.increment('packages.skipped', this.skipped.size);
+
+    if (this.skipped.size > 0) {
+      const clone = this.skipped.values();
+      this.skipped.clear();
+
       for (const job of clone) {
         this.changesConsumer?.unshift({ ...job, retry: 0 });
       }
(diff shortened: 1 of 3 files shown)


src/watch.ts
```diff
@@ -117,10 +124,29 @@ export class Watch {
     });
   }
 
+  /**
+   * Regurarly try to reprocess skipped packages.
```
Haroenv, line 128:
```suggestion
   * Regularly try to reprocess skipped packages.
```

src/watch.ts
```diff
@@ -117,10 +124,29 @@ export class Watch {
...
+    if (this.skipped.length > 0) {
+      const clone = [...this.skipped];
+      for (const job of clone) {
+        this.changesConsumer?.unshift({ ...job, retry: 0 });
+      }
+    }
+
+    setTimeout(() => {
+      this.checkSkipped();
+    }, 60000);
```
Haroenv, line 142:
```suggestion
    }, config.retrySkipped);
```

bodinsamuel:
You were too fast @Haroenv I wasn't done 😬

src/watch.ts
```diff
@@ -47,6 +52,14 @@ export class Watch {
    *      until an other package is updated.
    *      It will never be up to date because he receive event at the same pace
    *      as they arrive in listener A, even if it's not the same package.
+   *
+   * --- Retry strategy.
+   *  Each packages update needs to be processed in order so there is a first naive strategy that will block the queue.
+   *  It will retry the same package after an exponential backoff, N times.
+   *
+   *  After N times, this update will be discarded in the this.skipped
+   *  This Map will be regularly reprocessed to avoid loosing jobs.
```
Haroenv, line 61:
```suggestion
   *  This Map will be regularly reprocessed to avoid losing jobs.
```

src/watch.ts
```diff
@@ -47,6 +52,14 @@ export class Watch {
...
+   *  This is an in-memory only retry, if the process is stopped, skipped job are lost.
```
Haroenv, line 62: maybe this should be persisted somewhere? Index settings? Or would you rely on bootstrap (very infrequent) for that

src/watch.ts
```diff
@@ -129,29 +166,25 @@ export class Watch {
...
-        // Delete package directly in index
-        // Filter does not support async/await but there is no concurrency issue with this
-        throw new Error('deleted');
-      }
-      const res = await npm.getDoc(change.id, change.changes[0].rev);
+    if (job.retry > 0) {
+      // retry backoff
+      const backoff = Math.pow(job.retry + 1, config.retryBackoffPow) * 1000;
+      log.info('Retrying (', job.retry, '), waiting for', backoff);
+      await wait(backoff);
```
Haroenv, line 173: does this slow down the processing of other items, or is it just this one promise?

src/watch.ts
```diff
@@ -129,29 +166,25 @@ export class Watch {
...
+    }
 
-      if (isFailure(res)) {
-        log.error('Got an error', res.error);
-        return;
-      }
+    if (change.deleted) {
+      // Delete package directly in index
```
Haroenv, line 177:
```suggestion
      // changesConsumer deletes the package directly in the index
```

Haroenv (approved):
let's try this

src/watch.ts
bodinsamuel: Yeah, probably the next step. Two solutions: 
- a persistent DB but that would require more infra, making this project a bit more complexe to manage. eg: redis in dev vs redis in prod, means docker-compose + a redis instance in GCP + monitoring etc...
- using a dedicated algolia index. While algolia is not perfect for queuing, it is a simple and effective solution if we don't perform too much operation on it.

src/watch.ts
bodinsamuel: As mentioned in the comments above, it is in fact blocking because:
- if it's third party API failure, this is going to impact all others packages so no need to rush anything
- if it's only related to this package, at worst we have waited 1minutes.

Since we are have not achieved proper real time, I feel like it's okay. Wdyt?

src/watch.ts
Haroenv: I'm ok with going for the simpler (algolia) solution, as it doesn't require provisioning new environments and stays easy to run locally

PR #819, opened 2021-12-12
feat: refresh packages

bodinsamuel:
Introduce the long-awaited live package refresh that will avoid the need for full bootstrap (most of the time).

The strategy is to store the `rev` (the internal version of the package) in the record.
The `expiresAt` facets is queried, sorted and picked by oldest. 
We query some records (batch of 20 every 2 minutes, to avoid filling the queue) from the oldest facet value and update them like regular job. 

To completely enable this feature we need the `rev` so for the moment it's not possible and will require one more bootstrap.
I don't know yet if it's enough to process everything in a timely fashion.


src/watch.ts
```diff
@@ -153,6 +162,70 @@ export class Watch {
...
+    const list = Object.keys(res.facets!['_searchInternal.expiresAt']).sort();
+    log.info(' > Found', list.length, 'expiration values');
+    if (list.length <= 0) {
+      return;
+    }
+
+    const pick = list.shift()!;
+    const expiresAt = new Date(parseInt(pick, 10));
+    if (expiresAt.getTime() > Date.now()) {
+      log.info(' > Oldest date is in the futur');
```
Haroenv, line 192:
```suggestion
      log.info(' > Oldest date is in the future');
```

src/watch.ts
```diff
@@ -153,6 +162,70 @@ export class Watch {
...
+    log.info(' > Found', pkgs.hits.length, 'expired packages');
+
+    const pushed: string[] = [];
+    for (const pkg of pkgs.hits) {
+      if (!pkg.rev) {
+        continue;
+      }
+      pushed.push(pkg.objectID);
+
+      // If an update come at the same time, this could override the update with old rev.
```
Haroenv, line 212: do you mean while this function is called, or right after this function is called? Could we use a semaphore, and delay adding items to changes consumer until this process is finished, to make sure it's first? Or maybe read the changes that happen between start of this process and end of this process, and skip adding if it's already refreshed for another reason?

Haroenv (commented):
that's a good idea!

bodinsamuel:
I'll merge, I need those update to speedup watch.ts

PR #951, opened 2022-04-25
feat(ranking): only boost actually exact matches

Haroenv:
Some examples where the previous filter worked better:

places -> places.js is a better first match than places, but the other results are fairly relevant (except the babel one)

Places only exact match works better
- react (removes re-act and reactjs which are bad packages)
- bootstrap (removes unpopular bootstrapjs package)

You can check these changes out in the npm-search_old index where I've tried out the change

cc @MartinKolarik

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -10,7 +10,8 @@ const indexSettings: Settings = {
     'owners.name',
   ],
   attributesForFaceting: [
-    'filterOnly(_searchInternal.alternativeNames)' /* optionalFacetFilters to boost the name */,
+    'name' /* optional filter for exact matching */,
+    'filterOnly(_searchInternal.alternativeNames)',
     'filterOnly(bin)',
     'searchable(keywords)',
     'searchable(computedKeywords)',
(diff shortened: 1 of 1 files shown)


bodinsamuel (approved):

Haroenv:
There's a couple queries from https://github.com/algolia/npm-search/pull/194 (the original PR) which have worse results now:

- reactnative (doesn't show the actual react-native package, as it's not actually as popular as tools that are used in emotion etc.)
- react native

I don't really see a solution for that though. React native is popular as a concept, but less popular than tools used everywhere that happen to have react native support, especially as the match is also in the name.
The other

MartinKolarik:
Overall this is probably better than before (and I know it's easier), but I didn't want to go this route because the number of legit packages that use `.js` in their name isn't actually so small and some of them are even among the most popular projects at jsDelivr. From the top 100 packages: hls.js, popper.js, video.js, instantsearch.js, videojs-vtt.js, ua-parser-js, autocomplete.js, chart.js, mescroll.js, mux.js and I can think of quite a few off the top of my head too.

Haroenv:
I see, I didn't think a reindex was the right solution here, as that takes a really long time, but I wonder if some combination of `popular` / `popularName` and `alternativeNames` could be used 🤔 We indeed still want to rank those "obvious" packages right ahead of a very similar one without the .js (that could be second position)

MartinKolarik:
Yes, I'd be ok if you always got the exact match first and then the other version second, even if the second was the better one, but this change makes it a lot worse for the case where `.js` is the correct one, particularly when the name also makes sense as a keyword, e.g.:

image

MartinKolarik:
Indeed, I mixed up the naming. Still, does that mean we'll need a reindex?

Haroenv:
I can’t think of a solution without reindex for now

Issue #941, opened 2022-04-22
Improve ranking


MartinKolarik:
After including package downloads within the search results, we noticed a very common problem with relevance. I believe it's caused by ignoring "js" and probably typo tolerance, and I wonder if there's a way to improve this. A few examples:

1. https://www-jsdelivr-com-pr-400.onrender.com/?query=vue
First match "vue" (correct), then "vue.js" (garbage), then "vuejs" (garbage, even marked as deprecated), then relevant vue-related packages.

2. https://www-jsdelivr-com-pr-400.onrender.com/?query=react
First match "react" (correct), then "reactjs" (garbage), then re-act (garbage), then relevant react-related packages.

3. https://www-jsdelivr-com-pr-400.onrender.com/?query=bootstrap
First match "bootstrap" (correct), then "bootstrapjs" (garbage).

I know that in some cases, it's the other way around and "package.js" is the correct popular name while "package" is garbage, so this isn't easy, but at the same time, npm has *very* often multiple packages with similar names, so this happens for the majority of searches. To get correct results, we'd need to allow the "relaxed" matching (js ignore and typos) only if a popular match isn't found.

I'm not sure how doable this is but one idea I got so far - maybe instead of ignoring "js" globally, only set the other version as alternate name for packages that seem relevant (downloads above some threshold, not deprecated, have readme, ...). Non-relevant packages wouldn't get this relaxed matching.

Haroenv:
Technically I'm not sure how this could be done, but we could indeed have a certain download number threshold before adding alternative names maybe? That way at least those three examples would not show up anymore, it bothered me too.

bodinsamuel:
just ordering by download should help no?

Haroenv:
The problem is that custom ranking comes after word relevance, unless you mean resorting in the frontend @bodinsamuel

MartinKolarik:
Downloads are already used as part of the ranking but as I understand they have lower priority than the match on project name, so they don't help here.

Something I'm not 100% sure about though - I would expect that "ignored words" apply only to whole words, not parts of them so in "vuejs" the "js" suffix would not be ignored. But based on the search results, it either is ignored even when it's a suffix (in which case my proposed solution would help) or it's some other algolia feature that makes this rank higher (in which case the threshold idea may not help).

Haroenv:
The js is not an ignored word, but it's listed as an alternativeName, which is listed after the official name

MartinKolarik:
[...]

Issue #657, opened 2021-07-12
Ignore some packages ?


bodinsamuel:
Just a proposal, I'm not even convinced this is a good idea.
The only thing I want to do is gaining some time in the indexing, and the packages that are the less relevant are usually the slowest because there is no cache anywhere.
We could:
- exclude them entirely
- add them without additional data, to the index with a flag
- process them like every other with a flag
- do nothing

# Security holded
e.g: https://www.npmjs.com/package/kamonetucbp
They do not provide any value and cache miss everywhere obviously.
  
# Test user and test packages
I don't have a very exhaustive list, but for example Ryan (ts maintainer) has created a lot of package to test DefTyped release.
https://www.npmjs.com/~ryancavanaugh

I also saw a lot of tests and repetitive pattern, but not sure how to catch them without ignoring others too.

# Old packages

Probably not a good idea to ignore old packages (> 6 years old), some well appreciated packages do not receive update often.

Haroenv:
For packages that are security held, we can indeed probably bail out on most processing, and stop after formatPkg, possibly with a special flag  in searchInternal to mention that we stopped processing.

For older packages, I wonder if we could do a cache ourselves (aka the bootstrap index)? However this means downloads for example don't get updated, which I don't think is a good idea 🤔

MartinKolarik:
I was just about to file an issue for the security holding packages since it's also related to #941. Those don't have any extra flags (compared to e.g. deprecated) so they don't get any search penalization. I would say that if a package points to https://github.com/npm/security-holder/, it should get a flag and be pushed down in the search results.

Haroenv:
good call @MartinKolarik, it can be marked as `deprecated: "held by security"` so that it has the same penalty as a deprecated package. Maybe it can also have the `alternativeNames` not calculated so you're not likely to receive it. Would you be willing to make a PR?

MartinKolarik:
@Haroenv sure. With the alternative names condition we're getting to the #951 territory so I can add the other conditions too (e.g. downloads) and make a PR for overall relevance improvement.

bodinsamuel:
A dedicated boolean would be great ☺️

MartinKolarik:
So a separate boolean or `deprecated` with explanation in `deprecatedReason` as @Haroenv suggested? Boolean seems cleaner but we'll have to add it to index config too then.

Haroenv:
either is fine really, dedicated key makes sense indeed. Maybe in _searchMetadata

Issue #930, opened 2022-04-14
> 5 days delay in indexing


MartinKolarik:
We got a report today about a package missing from the search results (https://github.com/jsdelivr/jsdelivr/issues/18387), and I see the first release was five days ago. There was also a similar issue recently (#928). Is there anything we could do to reduce the delay?

bodinsamuel:
yep indeed. 
It seems NPM bumped their sequence again and we are 60K updates behind. 
Not so much we can do since we need to process update sequentially.
The same issue all over again :'( 

We had lost metrics monitoring since a few weeks (and had no time to look up), I have added a more annoying alerting so that I don't skip those. 

--- 

For JsDelivr we should really sync and find not an alternative but a decent way to fallback. Could be with a secondary real time index that has no additional info combined with an internal API to refresh cache.

MartinKolarik:
Do you know what's slowing down the indexing most? I'm wondering how the fallback would work because if you visit a page like https://www.jsdelivr.com/package/npm/quantdom it's easy, but to get there, you'd usually search at https://www.jsdelivr.com/?query=quantdom and if that returns some results, there's no way for us to know we need to fallback.

At the same time, we use jsDelivrHits for sorting the search results and also plan to add an option for filtering based on moduleTypes/styleTypes fields, so using a lighter index *always* is not an option. Would it make sense to maybe insert packages to the main index immediately, with some kind of "in progress" flag, and update it later?

Haroenv:
The index is back up to date now, thanks for standing by!

PR #965, opened 2022-05-16
feat(ranking): add popularAlternativeNames, detect security packages (#941)

MartinKolarik:
1. Adds a new `isSecurityHeld` flag as discussed in #657. I went with a property at the same level as `isDeprecated`(and not `_searchInternal`) since it may be interesting for the end users too.
2. The property is then used for #941 along with some other checks. Popular packages are excluded from the checks since I think those should get the alternative names even if they are deprecated (applies to e.g. [request](https://www-jsdelivr-com-pr-400.onrender.com/?query=request)).


MartinKolarik:
Note that I focused on the ranking aspect here and didn't add any skipping logic for the security packages - can add that too if you want.

src/config.ts
```diff
@@ -179,6 +181,8 @@ export const config = {
   retrySkipped: ms('1 minute'),
   retryBackoffPow: 3,
   refreshPeriod: ms('2 minutes'),
+  alternativeNamesNpmDownloadsThreshold: 5000,
+  alternativeNamesJsDelivrHitsThreshold: 10000,
```
Haroenv, line 185: the jsDelivr threshold seems quite high, wouldn't this rule out eg. create-react-app, which only has 950 downloads on jsDelivr

src/saveDocs.ts
```diff
@@ -58,6 +59,18 @@ async function addMetaData(pkg: RawPkg): Promise<FinalPkg> {
...
 
+  const keepAlternativeNames =
+    final.popular ||
+    (!final.isDeprecated &&
+      !final.isSecurityHeld &&
+      (final.downloadsLast30Days >=
+        config.alternativeNamesNpmDownloadsThreshold ||
+        final.jsDelivrHits >= config.alternativeNamesJsDelivrHitsThreshold));
+
+  if (!keepAlternativeNames) {
```
Haroenv, line 70:
```suggestion
  const hasFewDownloads =
    final.downloadsLast30Days <= config.alternativeNamesNpmDownloadsThreshold &&
    final.jsDelivrHits <= config.alternativeNamesJsDelivrHitsThreshold;

  const dropAlternativeNames =
    !final.popular &&
    (final.isDeprecated || final.isSecurityHeld || hasFewDownloads);

  if (dropAlternativeNames) {
```

this reads a little simpler to me, what do you think?

Haroenv (commented):
interesting approach!

Haroenv:
I'd like to see what impact that has on the index, but not sure how to test it without doing a complete index, do you have any idea?

Haroenv:
Maybe what we could do is instead of removing alternative names completely, we add a new key called popularAlternativeNames in the searchInternal, add the query rule to that instead of alternative names, make sure the original name Is still in popularAlternativeNames (maybe?). That way after a reindex, we can compare the quality between popularAlternativeNames and alternativeNames

MartinKolarik:
> Maybe what we could do is instead of removing alternative names completely, we add a new key called popularAlternativeNames in the searchInternal, add the query rule to that instead of alternative names, make sure the original name Is still in popularAlternativeNames (maybe?). That way after a reindex, we can compare the quality between popularAlternativeNames and alternativeNames

Agree, that's a good idea.

src/@types/pkg.ts
```diff
@@ -48,6 +49,7 @@ export interface RawPkg {
   downloadsLast30Days: number;
   downloadsRatio: number;
   humanDownloadsLast30Days: string;
+  jsDelivrHits: number;
```
bodinsamuel, line 52: indeed good catch

bodinsamuel (approved):
LGTM
This will require full reindex unfortunately. I need to finish the work on incremental full reindex that was started #819

MartinKolarik:
@bodinsamuel unrelated but do you know what's currently the bottleneck in full reindex? E.g. would it help to run it on a larger VM with higher concurrency?

MartinKolarik:
Also in issues like #930, I was wondering if throwing more resources at it would help as that's something we maybe would be able to do.

MartinKolarik:
> Rate limit (429) from NPM registry
We should get in touch with them to see what can be done

That's interesting as our services query npm a lot too and we don't have this issue. Is that the registry or some other service? (maybe the downloads API?). We previously had a similar problem with GitHub and it the limit was too low-level for them to do anything about it, we solved it by assigning multiple IPs to our VMs and distributing the outgoing requests across those (but that's probably not possible at PaaS).

> We can barely parallelise work

I checked the code and it indeed looks problematic but I think there might be some not too hard solutions. Process N packages at once, and track which are being processed. When a second event for the same package arrives, either put it in an intermediate queue (which has some small size limit) or simply pause until the previous event for the package competes.

MartinKolarik:
> E.g: For package A and package B, if I receive A1, B1, A2, for whatever reason I process B1 first, I can not say "the process is at B1". If the process crashes during this time, I need to be able to say "I processed B2 but not A1 and A2".

I may be missing something but not sure how the concurrency is relevant here - the inserts/updates here should be idempotent so I would just ignore the fact that B1 was processed and say "I'm at whatever was before A1" - which makes this part the same as for serial processing.

> It's a bit awful how simple it seems at first and then realising that it actually requires more complexity to be able to process things correctly.

I'm not surprised at all 😆

bodinsamuel:
> should be idempotent so I would just ignore the fact that B1 was processed and say "I'm at whatever was before A1"

Yeah my explanation was incomplete, let me try again. With this sequence of updates:
```sh
sequence 1: A1 (update A)
sequence 2: B1 (create B)
sequence 3: A2 (update A)
sequence 4: B2 (delete B)
```

We start at `sequence 0`, we process B1 and B2 correctly but A1 is blocking for whatever reason. The process needs to store the progress of the sequence somewhere to be able to start where it left off in case of crash or restart.
- If I store `sequence 4` I skipped 2 updates
- If I store `sequence 0` I redo 4 updates

As you mention, updates are now idempotent (which was not the case before), so it's fine to reprocess until you realise that if you already processed `sequence 4` you deleted the package B. So by reprocessing all sequences you endup actually recreating the package B. (or in other scenario you virtually rollback to a previous version)

Note:
- it's more than 4 sequences, could be hundred or thousand. So could impact a lot of packages.
- we could say we don't care and it's an acceptable downside
- we could add one more check that would get the last update date for a package and skip all updates where `sequence.updateAt < current.updateAt` but it needs to be stored somewhere and Algolia can't be relied upon because search is only eventually consistent.

That's why I would favor having a DB or a queue, to store all updates and redo only the ones that were not processed.
I hope the explanation is bit clearer :p

MartinKolarik:
> So by reprocessing all sequences you endup actually recreating the package B. (or in other scenario you virtually rollback to a previous version)

This seems acceptable because the window for this error is only as big as the concurrency level so very soon you delete B again. You could even say that from an outside observer's perspective, you wouldn't get to `seq 4` before the crash in the first place with serial processing, so the final correct state may still be achieved faster even if it's `B1 -> B2 -> B1 -> B2` instead of `B1 -> B2`

Anyway, an external queue definitely makes sense here, just saying that if an easier solution was more desirable,  it could probably still work.

MartinKolarik:
Changed to `popularAlternativeNames` and `alternativeNames ` are kept as they were.

MartinKolarik:
> Note that I focused on the ranking aspect here and didn't add any skipping logic for the security packages - can add that too if you want.

Added now.

src/saveDocs.ts
```diff
@@ -58,6 +63,18 @@ async function addMetaData(pkg: RawPkg): Promise<FinalPkg> {
...
+  const hasFewDownloads =
+    final.downloadsLast30Days <= config.alternativeNamesNpmDownloadsThreshold &&
+    final.jsDelivrHits <= config.alternativeNamesJsDelivrHitsThreshold;
+
+  const addPopularAlternativeNames =
+    final.popular ||
+    (!final.isDeprecated && !final.isSecurityHeld && !hasFewDownloads);
+
+  if (addPopularAlternativeNames) {
+    final._searchInternal.popularAlternativeNames = [];
```
Haroenv, line 75:
```suggestion
    final._searchInternal.popularAlternativeNames = final._searchInternal.alternativeNames;
```

isn't this meant to be something like this?

Haroenv, line 75: We also can do the condition before creating final:

```js
const popularAlternativeNames = [];

const hasFewDownloads =
  final.downloadsLast30Days <= config.alternativeNamesNpmDownloadsThreshold &&
  final.jsDelivrHits <= config.alternativeNamesJsDelivrHitsThreshold;

const addPopularAlternativeNames =
  final.popular ||
  (!final.isDeprecated && !final.isSecurityHeld && !hasFewDownloads);

if (addPopularAlternativeNames) {
  popularAlternativeNames.push(...pkg._searchInternal.alternativeNames);
}

const final = {
  // ...
  _searchInternal: {
    ...pkg._searchInternal,
    popularAlternativeNames,
  },
};
```

or in a function:

```js
function s() {
  const popularAlternativeNames = getPopularAlternativeNames(pkg);
  const final = {
    // ...
    _searchInternal: {
      ...pkg._searchInternal,
      popularAlternativeNames,
    },
  };
}

function getPopularAlternativeNames(pkg) {
  const hasFewDownloads =
    pkg.downloadsLast30Days <= config.alternativeNamesNpmDownloadsThreshold &&
    pkg.jsDelivrHits <= config.alternativeNamesJsDelivrHitsThreshold;

  const addPopularAlternativeNames =
    pkg.popular ||
    (!pkg.isDeprecated && !pkg.isSecurityHeld && !hasFewDownloads);

  if (addPopularAlternativeNames) {
    return pkg._searchInternal.alternativeNames;
  }

  return [];
}
```

MartinKolarik: Lol yes, need to update the tests too.

MartinKolarik: We could but then each property is on a different object and some null checks are needed, so it looks like this:

```js
  const popularAlternativeNames: string[] = [];

  const hasFewDownloads =
      (download?.downloadsLast30Days || 0) <= config.alternativeNamesNpmDownloadsThreshold &&
      hit.jsDelivrHits <= config.alternativeNamesJsDelivrHitsThreshold;

  const addPopularAlternativeNames =
      download?.popular ||
      (!pkg.isDeprecated && !pkg.isSecurityHeld && !hasFewDownloads);

  if (addPopularAlternativeNames) {
    popularAlternativeNames.push(...pkg._searchInternal.alternativeNames);
  }
```

Haroenv: fine as-is then 👍

MartinKolarik:
Then I think it's all done here.

src/config.ts
```diff
@@ -85,14 +87,14 @@ const indexRules: Rule[] = [
     objectID: 'promote-exact',
     description: 'promote exact matches',
     condition: {
-      pattern: '{facet:_searchInternal.alternativeNames}',
+      pattern: '{facet:_searchInternal.popularAlternativeNames}',
```
Haroenv, line 90: we should change the query rule only after the reindex is completed, otherwise this will automatically change everything.

MartinKolarik: True, reverted everything to `alternativeNames` for now.

Haroenv (approved):

bodinsamuel:
Well I don't have admin right neither and we do have the the "run fork pr" in circleci, it should run 🤔

Haroenv:
merged in #975

PR #1074, opened 2022-11-01
fix(config): update jsDelivrHitsEndpoint url

MartinKolarik:
The current URL uses an undocumented feature we'll have to remove soon due to path conflicts with some new endpoints. To avoid any disruption here, this PR switches to a new URL on our preview environment.

Once we ship all changes to the production env (and this new endpoint is added and the old one removed), I'll send another PR to switch from `data-jsdelivr-com-preview.onrender.com` to `data.jsdelivr.com`

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -156,7 +156,7 @@ export const config = {
   npmDownloadsEndpoint: 'https://api.npmjs.org/downloads',
   npmRootEndpoint: 'https://registry.npmjs.org',
   jsDelivrHitsEndpoint:
-    'https://data.jsdelivr.com/v1/stats/packages/npm/month/all',
+    'https://data-jsdelivr-com-preview.onrender.com/v1/stats/packages/all?period=month&type=npm',
   jsDelivrPackageEndpoint: 'https://data.jsdelivr.com/v1/package/npm',
   typescriptTypesIndex:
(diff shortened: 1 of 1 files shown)


MartinKolarik:
This change also makes the memory footprint of the response about 20x smaller by excluding packages with very few hits (< 100 / month) on the server, as they are being filtered out here anyway.

Haroenv (approved):

MartinKolarik:
Ping 😄

Haroenv:
Sorry, I lost access to merge on this repository and haven't yet received a response from security. Will update

Haroenv:
Ok I have access again 🙌 Should this PR still be merged now?

Haroenv:
@MartinKolarik it looks like the test fails and jsDelivrHits is significantly lower than expected for some packages. Do you think that's expected?

MartinKolarik:
Hmm what tests exactly?

MartinKolarik:
Oh, I see. That's the number of packages in the response. That's affected by this:

> This change also makes the memory footprint of the response about 20x smaller by excluding packages with very few hits (< 100 / month) on the server, as they are being filtered out here anyway.

So yes, it's expected the number is much lower now, I didn't realize there was a test for it.

Haroenv:
Thanks, I've updated it in d99c31f

Issue #1089, opened 2022-12-05
`getObject()` for Yup reports types of DefinitelyTyped


karlhorky:
Hi @Haroenv and other maintainers 👋 Hope you are well! Thanks again for your continued effort on the projects in the Algolia ecosystem!

I'm not sure if I'm reporting this in the right place, but I wanted to reach out about what appears to be a data problem with `npm-search` index (we've only seen a single library so far):

Running `index.getObject()` with the package name [`yup`](https://www.npmjs.com/package/yup) returns the following:

```js
{
  types: {
    ts: 'definitely-typed',
    definitelyTyped: '@types/yup',
  },
}
```

However, on the npm website, [the `yup` package](https://www.npmjs.com/package/yup) is reported as having built-in types and [the `@types/yup` package](https://www.npmjs.com/package/@types/yup) has been deprecated:

<img width="364" alt="Screenshot 2022-12-05 at 19 58 19" src="https://user-images.githubusercontent.com/1935696/205720370-7f1ab291-c5c9-4978-b0d3-b39a76cbc0f3.png">

<img width="483" alt="Screenshot 2022-12-05 at 19 59 46" src="https://user-images.githubusercontent.com/1935696/205720616-4f12161f-f5a0-4880-9fd0-5f2866428801.png">

Both the current version and the beta version of `yup` have no `"types"` or `"typings"` keys in the `package.json`:

- https://unpkg.com/browse/yup@0.32.11/package.json
- https://unpkg.com/browse/yup@1.0.0-beta.8/package.json

For the current `yup@0.32.11`, there is however no `index.d.ts` in the root:

- https://unpkg.com/browse/yup@0.32.11/index.d.ts (404)

But there is one in `yup@1.0.0-beta.8`:

- https://unpkg.com/browse/yup@1.0.0-beta.8/index.d.ts

Is the npm website or Algolia's index wrong? (seems like npm website is wrong 🤔)

---

Our usage:

```ts
const client = algoliasearch(
  'XXX',
  'xxxxxxxxxxx',
);

const index = client.initIndex('npm-search');

const results = await index.getObject<AlgoliaObj>(dependency, {
  attributesToRetrieve: ['types'],
});

const definitelyTypedPackageName = results.types?.definitelyTyped;
```

https://github.com/upleveled/preflight/blob/5af8a398ec370c79e22ab770307ec81d574c8e29/src/checks/noDependencyProblems/noDependenciesWithoutTypes.ts#L63-L67

[...]

PR #1091, opened 2022-12-07
Fix bug that identify wrongly some TS packages as Definetely Typed 

Josehower:
This PR solves the issue that appears to be a data problem with `npm-search` index listing yup (we've only seen a single library so far) as non-TS library.

To solve it, applies the suggestion by @jablko and reverse the logic in the `getTypeScriptSupport` function, preventing yup, or other libraries written in TypeScript from being flagged as definitely typed even if an outdated/deprecated Definitely Typed reference exists.

closes https://github.com/algolia/npm-search/issues/1089

### DIFF
--- a/src/typescript/index.ts
+++ b/src/typescript/index.ts
@@ -86,7 +86,6 @@ export function getTypeScriptSupport(
         },
       };
     }
-
     datadog.increment('jsdelivr.getTSSupport.miss');
 
     return { types: { ts: false } };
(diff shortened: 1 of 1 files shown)


Haroenv:
Thanks! This fix won't automatically be taken in account, but for new/updated packages it will work already. I'll launch a full reindex as well parallel soon, but that will take a long time

PR #1107, opened 2023-02-11
fix(config): update jsDelivrHitsEndpoint url

MartinKolarik:
Following up on #1074 and switching the URL to our prod env.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -156,7 +156,7 @@ export const config = {
   npmDownloadsEndpoint: 'https://api.npmjs.org/downloads',
   npmRootEndpoint: 'https://registry.npmjs.org',
   jsDelivrHitsEndpoint:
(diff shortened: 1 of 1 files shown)


Haroenv:
Looks like something went wrong with CI, I'm not sure if it will build, but I'll try to fix it

MartinKolarik:
Can you confirm if/when this is live? In our APM, I'm seeing only requests coming from version 1.7.7, which is quite old 🤔 no signs of either 1.7.15. or 1.7.14.

Haroenv:
hm, maybe something didn't work in the release script. I'm seeing a failure with docker

MartinKolarik:
Any idea when you might be able to fix it and deploy the latest version? Since we thought the previous changes were live, we've already deployed our API changes and the URI used here to get package stats no longer works.

PR #1143, opened 2023-10-20
fix(ci): Trigger CI checks on branch push, not only on PR creation

pixelastic:
I'm updating the `validate` workflow to trigger on each push, not only when a PR is created.

This is how the config on CircleCI used to work and allow to check the status of the checks before actually creating a PR (thus avoiding noise as one can debug checks without creating a PR for that)


pixelastic:
I'm also taking this PR as the opportunity to check/debug the publish workflow on master again

PR #1147, opened 2023-10-21
ci: allow to run on pull request

bodinsamuel:
Renovate is excited because everything is green, but it's only because the CI didn't run :D


pixelastic:
Thanks!

I had assumed that putting `on.push` would have also run it on any PR, and not specifying a branch would have run it all the time.

PR #1154, opened 2023-10-24
ci: Move flaky npm tests into specific scheduled API Control script

pixelastic:
Debugging the CI/Renovate was made more complex because of the flakiness of the tests targeting npm. This PR moves those specific tests into their own file. They are not longer executed as part of the regular `yarn test` script, but the CI will still run them once per day, and report any failure as a new GitHub issue.

- The three tests (`findAll`, `getDoc` and `getInfo` from the npm module) have been moved to `__tests__/api-control/`
- This folder is excluded by default from all `yarn test` runs (option `testPathIgnorePatterns` from `jest.config.js`)
- `yarn test:api-control` can be called locally to only run those tests
- The `api-control.yml` workflow has been added to GitHub pages and will run every day at 10am. It will run `yarn test:api-control`, and if the test fails, will create a new issue in the repo with the details of the error and a link to the run.

I did test the workflow as much as I could on this PR, but like all things CI-related, the real test will happen once it's executed for real on the main branch.

We might want to move other similar tests to `__tests__/api-control`. I will also update `CONTRIBUTING.md` with updated documentation about those changes once the whole CI is fixed. 

I'm open to other names than "API Control", but that was the more explicit and short I could find. Naming is hard. Smoke test? Conformity test? Anti-Corruption test? Contract test? Probe test?


.github/workflows/api-control.yml
```diff
@@ -0,0 +1,25 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18.18'
+          cache: 'yarn'
+
+      - name: 📦 Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: 🛂 API Control
+        run: GITHUB_RUN_ID="${{ github.run_id }}" yarn test:api-control
```
pixelastic, line 25: `GITHUB_RUN_ID` is then used internally by the script as a condition to know if it's being run locally or from the CI

scripts/test-api-control
```diff
@@ -0,0 +1,50 @@
...
+		--testPathIgnorePatterns=''
+	exit $?
+fi
+
+# Running on CI, creating an issue on failure
+echo "Wait while we run the tests"
+output=$(jest \
+	./src/__tests__/api-control \
+	--forceExit \
+	--testPathIgnorePatterns='' 2>&1)
```
pixelastic, line 27: By default, jest redirects all its output to stderr. `2>&1` is here to redirect it back to stdout so we can capture it

Haroenv (approved):
having a retry could be interesting, but this is a good start already, thx!

PR #1161, opened 2023-10-27
chore: Update to TypeScript v5 as well as all relevant tooling

pixelastic:
The goal of this PR was to upgrade TypeScript to v5, as suggested by Renovate. Doing so triggered a serie of needed updates in the tooling (ESLint & Jest ). I'll detail them below as it will be easier to follow than comments on file changed.

I started by upgrading TypeScript from `4.7.4` to `5.2.2`.

## Linting

This upgrade broke the linting script with the error `WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.` So, I upgraded our ESlint TypeScript plugins (`@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser`) to their latest versions. 

Now, the linting fails with a more descriptive error:
```
Error: Error while loading rule '@typescript-eslint/prefer-optional-chain': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.
```

The new `@typescript-eslint` version now requires that we specify which `parser` and `parserOptions` to use for `.ts` files. I added the following config to `.eslintrc`:

```js
parser: '@typescript-eslint/parser',
parserOptions: {
  sourceType: 'module',
  project: './tsconfig.json',
},
```
At this stage, the linting process works, and output 53 errors. Most of them are along the lines of:

```
Definition for rule '@typescript-eslint/sort-type-union-intersection-members' was not found  @typescript-eslint/sort-type-union-intersection-members
```

This is because the rule `@typescript-eslint/sort-type-union-intersection-members` no longer exists (it has been renamed), and the `eslint-config-algolia` plugin we extend references it. So, in a very meta way, ESLint warned us that we are using a rule that does not exists in ESLint. I upgraded `eslint-config-algolia` to the latest version, where the rule has been correctly renamed to `@typescript-eslint/sort-type-constituents`.

We're down to 19 linting errors now, most along the lines of:

```
Parsing error: ESLint was configured to run on `<tsconfigRootDir>/src/typescript/index.test.ts` using `parserOptions.project`: <tsconfigRootDir>/tsconfig.json
However, that TSConfig does not include this file. Either:
- Change ESLint's list of included files to not include this file
- Change that TSConfig to include this file
- Create a new TSConfig that includes this file and include it in your parserOptions.project
```

This basically means that now that we've defined the TS parser and its options, it's warning us that we are linting a file that hasn't been included in the TS config. So, I updated `tsconfig.js` to replace `include": ["src/*.js", "package.json"]` with `include": ["src/**/*.ts", "package.json"]`. This reduced the number of errors to 9. (I'm not sure why `package.json` is here, but I left it).

Interesting thing to note, I have my IDE configured to format/fix the content of files on save, based on any linter/fixer configured in the project. It detected that we had Prettier, so it did format `tsconfig.json` through Prettier automatically, which had the side effect of putting short array definitions on one line, and add commas **after** the comments. That wasn't expected and I can revert it if needed, but I think it makes sense to format that file as well so I kept it.

At this stage, I still had the same linting error as above triggering on `elastic-apm-node.js`, `jest.config.js` and `release.config.js`. Those are JavaScript files, not TypeScript files, so it doesn't make sense to run TS-specific linter rules on them.

I updated `.eslintrc` to move all TS specific config into the TS override like so:

```json
  "overrides": [
    {
      "files": ["**/*.ts"],
      "extends": ["algolia", "algolia/jest", "algolia/typescript"],
      "parser": "@typescript-eslint/parser",
      "parserOptions": {
        "sourceType": "module",
        "project": "./tsconfig.json"
      },
      "rules": {
        "consistent-return": ["off"],
        "no-dupe-class-members": ["off"]
      }
    }
  ]
```

This get rid of the above error (TS-specific linting rules were only applied to TS files) as expected, but introduced new ones. Now I had a lot of `no-continue` and `require-await` errors. Those were supposed to be ignored as we turned them off in the top-level `rules` key of `.eslintrc`, so why were they appearing now? 

Unfortunately, my previous change of defining all TS config in its own `overrides` had been replacing the top-level `rules` key with the TS-specific one, instead of merging them. So I renamed `.eslintrc` to `.eslintrc.js`, stored the default rules in `defaultRules` and created two specific `overrides`, one for JS and one for TS, to clearly prevent configuration leaking. 

With that, I got to only 7 errors left, 4 of them fixable. So I ran `yarn lint --fix` and got down to 3. Most of the fixed errors were `@typescript-eslint/prefer-optional-chaining` and converted stuff like `(repository === null || !repository.host)` to `(!repository?.host)`. The last cases were some `eslint-disable` comments for rules that didn't apply.

The last errors were easy to fix:
- In `Prefetcher.ts` I had to hoist the private `stateManager` and `queueIndex`
- I had to rename `config.test.js` to `config.test.ts` as it was the only js file in the folder and was importing a `.ts` file

At this stage, no more linting issues \o/.

## Test

Running tests created a similar situation of warning and errors because of outdated dependencies, like `Version 5.2.2 of typescript installed has not been tested with ts-jest. If you're experiencing issues, consider using a supported version (>=4.3.0 <5.0.0-0). Please do not report issues in ts-jest if you are using unsupported versions.`

So I upgraded `jest` and `ts-jest` and got a warning about:
```
Define `ts-jest` config under `globals` is deprecated. Please do
transform: {
    <transform_regex>: ['ts-jest', { /* ts-jest config goes here in Jest */ }],
},
```

Turns out the `jest.config.js` format slightly changed and
```
transform: {
    '^.+\\.[jt]sx?$': 'ts-jest',
  },
  globals: {
    'ts-jest': {
      diagnostics: false,
      tsconfig: `tsconfig.json`,
    },
  },
```
needed to be merged into
```
transform: {
    '^.+\\.[jt]sx?$': [
      'ts-jest',
      {
        diagnostics: false,
        tsconfig: `tsconfig.json`,
      },
    ],
  },
```

Running the tests now generated dozens of error, but most were because of a change in the default format of Jest snapshots, that now uses the short notation (`{` and `[`) instead of the verbose one (`Object {` and `Array [`). As I didn't want to update all snapshots in that PR (I wasn't comfortable enough with the changes I made not breaking tests), I updated `jest.config.js` to specifically use the legacy format with:

```
 snapshotFormat: {
    escapeString: true,
    printBasicPrototype: true,
  },
```

And that made all the tests pass \o/

## Build

**tl;dr;**: `yarn build` still fails. As my TypeScript skills are not very good, I can't tell if those are real errors we need to fix in the code, or tooling that needs to be configured differently as above, so I'll need someone else to pick it up from here.


.eslintrc.js
```diff
@@ -0,0 +1,39 @@
...
+        sourceType: 'module',
+        project: './tsconfig.json',
+      },
+      rules: {
+        ...defaultRules,
+        'consistent-return': 'off',
+        'no-dupe-class-members': 'off',
+      },
+    },
+  ],
```
Haroenv, line 38:
```suggestion
  rules: defaultRules,
  extends: ['algolia', 'algolia/jest'],
  overrides: [
    {
      files: ['**/*.ts'],
      extends: ['algolia', 'algolia/jest', 'algolia/typescript'],
      parser: '@typescript-eslint/parser',
      parserOptions: {
        sourceType: 'module',
        project: './tsconfig.json',
      },
      rules: {
        'consistent-return': 'off',
        'no-dupe-class-members': 'off',
      },
    },
  ],
```

this works too, no? And which files are .js? do we need that config still?

Haroenv (commented):
looks like formatPkg test needs some assertions, and lint config can be simplified, but good otherwise I think

Haroenv (approved):

.eslintrc.js
```diff
@@ -0,0 +1,36 @@
...
+    {
+      files: ['**/*.ts'],
+      extends: ['algolia', 'algolia/jest', 'algolia/typescript'],
+      parser: '@typescript-eslint/parser',
+      parserOptions: {
+        sourceType: 'module',
+        project: './tsconfig.json',
+      },
+      rules: {
+        ...defaultRules,
```
bodinsamuel, line 30: I don't think it's necessary to reapply rules, overrides only override what is specified

bodinsamuel (commented):
💯

.eslintrc.js
Haroenv: ah yes I missed that, indeed not needed. Will fix

pixelastic: Are you sure about that? That was the issue I was having, the `overrides` was overriding the whole `rules` from the top level, that's why I had to move from `.json` to `.js`

Haroenv: no, it was because of the `extends`, 4df2833 is more "correct"

pixelastic: No, you're right, that works. I thought I had an issue with not being able to merge an override rule with the main one, but that might have been caused by something else. Current config is lean and clean

PR #1225, opened 2023-12-15
fix: add package name to all records

MartinKolarik:
We use the package name as `objectID` in the helper indexes, but `objectID` is not searchable, which makes it very hard to interact with the indexes via the UI (for debugging). This adds the package name as an additional searchable attribute to all the data indexes and queues.


Haroenv (approved):
do you want to add it to searchableAttributes in settings too?

MartinKolarik:
Since we don't set it explicitly on the helper indexes, everything is searchable by default.

Issue #1233, opened 2023-12-18
Fix docker publishing


MartinKolarik:
I fixed an issue with the script path in #1232 but now publishing fails with:

```
SemanticReleaseError: Cannot push to the Git repository.
    at default (file:///home/runner/work/npm-search/npm-search/node_modules/semantic-release/lib/get-error.js:6:10)
    at run (file:///home/runner/work/npm-search/npm-search/node_modules/semantic-release/index.js:101:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async default (file:///home/runner/work/npm-search/npm-search/node_modules/semantic-release/index.js:275:22)
    at async file:///home/runner/work/npm-search/npm-search/scripts/publish-check.mjs:[19](https://github.com/algolia/npm-search/actions/runs/7248615388/job/19744902325#step:5:20):25 {
  code: 'EGITNOPERMISSION',
  details: '**semantic-release** cannot push the version tag to the branch `master` on the remote Git repository with URL `https://x-access-token:[secure]@github.com/algolia/npm-search.git`.\n' +
    '\n' +
    'This can be caused by:\n' +
    ' - a misconfiguration of the [repositoryUrl](https://github.com/semantic-release/semantic-release/blob/master/docs/usage/configuration.md#repositoryurl) option\n' +
    ' - the repository being unavailable\n' +
    ' - or missing push permission for the user configured via the [Git credentials on your CI environment](https://github.com/semantic-release/semantic-release/blob/master/docs/usage/ci-configuration.md#authentication)',
  semanticRelease: true
}
```

I believe this might be related to branch protection rules, but don't have high enough access to see those, so not sure. Also, the whole CI job ends without reporting errors even though it fails, which isn't great: https://github.com/algolia/npm-search/actions/runs/7248615388/job/19744902325

@pixelastic can you take a look please?

pixelastic:
Thanks for noticing. I'll have a look shortly. 

Doesn't seem to be related to docker per-se, it's the branching pre-check that validates if we need to do a release or not. I don't understand why it would need to push to the repo, I'll investigate.

MartinKolarik:
> I don't understand why it would need to push to the repo, I'll investigate.

For this I think: https://github.com/algolia/npm-search/commits?author=semantic-release-bot

pixelastic:
Issue should now be fixed.

[...]

PR #1234, opened 2023-12-19
Differentiate between publishing success / fail / skip

pixelastic:
This is the first part in fixing #1233 : being able to differentiate between a legitimate error (what's currently happening, with the token issue) and a normal skipped publishing (because there are no relevant changes to deploy).

I want to merge that on master to see if it now correctly mark as error because of the token issue.

(Note: I'll keep working on the second part on Thursday)


.github/workflows/publish.yml
```diff
@@ -15,9 +15,11 @@ jobs:
     needs: validate
     runs-on: ubuntu-latest
     permissions:
+      actions: write
```
pixelastic, line 18: This is needed to cancel the GitHub Action run (in case there are no publishing to actually do)

.github/workflows/publish.yml
```diff
@@ -34,13 +36,16 @@ jobs:
 
       - name: ❓ Check if a new version needs to be published
         id: publish-check
-        continue-on-error: true
-        run: yarn publish:check
+        run: echo "SHOULD_PUBLISH=$(yarn publish:check)" >> "$GITHUB_OUTPUT"
```
pixelastic, line 39: Previous version was using the exit code of `yarn publish:check` to define if a real publishing was needed or not. This worked, but this was swallowing any legitimate error in the `yarn publish:check` script, and considered them as normal behavior.

I'm now using GitHub Actions output (see: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter) to pass values from step to step.

Here I define an output called `SHOULD_PUBLISH`, with a value returned by `yarn publish:check` (that will be either `yes` or `no`), and I'll use that later to conditionally execute some steps.

.github/workflows/publish.yml
```diff
@@ -34,13 +36,16 @@ jobs:
...
+
+      - name: ⏹️ Cancel run
```
pixelastic, line 41: This step is technically not required, as if the `SHOULD_PUBLISH` is `yes`, then only the other two steps will be triggered.

But calling `gh run cancel` will mark the workflow as cancelled instead of succeeding, which is more readable in the UI
image

MartinKolarik (approved):

Haroenv (approved):

PR #1352, opened 2025-06-04
feat: alternative names for css packages

MartinKolarik:
Improves ranking for `tailwindcss` and similar using the same logic that was used for `js` suffixes.

### DIFF
--- a/src/__tests__/formatPkg.test.ts
+++ b/src/__tests__/formatPkg.test.ts
@@ -462,6 +462,36 @@ describe('alternative names', () => {
     `);
   });
 
+  it('name ending in .css', () => {
+    const pkg: GetPackage = {
+      ...BASE,
+      name: 'animate.css',
+    };
+    expect(formatPkg(pkg)!._searchInternal.alternativeNames)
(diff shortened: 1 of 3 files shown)


src/__tests__/formatPkg.test.ts
```diff
@@ -462,6 +462,36 @@ describe('alternative names', () => {
...
+
+  it('name ending in css', () => {
+    const pkg: GetPackage = {
+      ...BASE,
+      name: 'tailwindcss',
+    };
+    expect(formatPkg(pkg)!._searchInternal.alternativeNames)
+      .toMatchInlineSnapshot(`
+      Array [
+        "tailwindcss",
```
Haroenv, line 489: seems like we're a little inconsistent and don't create the `tailwind css` name, but that's probably ok

Haroenv (approved):
thanks, this looks like a great addition!

src/__tests__/formatPkg.test.ts
MartinKolarik: Yes, but that matches the existing `js` behavior

PR #1357, opened 2026-04-13
feat: allow `owners.name` to be facetable

jonchurch:
There is no reliable way to filter packages down to ones maintained by a given npm user. 

The current `owner` field is, as called out by the docs, the github user not the actual npm user. Which is surprising! 

This change would allow filtering by the actual npm user who is a maintainer (aka appears in the maintainers array, which is the closest public concept of an "owner" as npm exposes). 

The PR itself doesnt change the semantics of the currwnt `owner` field, just makes it so folks can filter to the `ownerS` array. 

See npmx.dev related issue: https://github.com/npmx-dev/npmx.dev/issues/2504

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -15,6 +15,7 @@ const indexSettings: Settings = {
     'searchable(keywords)',
     'searchable(computedKeywords)',
     'searchable(owner.name)',
+    'searchable(owners.name)',
     '_oneTimeDataToUpdateAt',
     '_periodicDataUpdatedAt',
     'deprecated',


MartinKolarik:
@pixelastic wdyt? Seems useful and we even discussed it [long ago](https://github.com/jsdelivr/www.jsdelivr.com/issues/140#issuecomment-348694478), but never added it.

Haroenv (approved):

jonchurch:
Hey awesome! Looks like it's working now as well, is that correct for the deployment model here?

I curl'd and it seems to work, so Im guessing that's that? thanks!

Haroenv:
Yes, the deploy has happened since

MartinKolarik:
@Haroenv I'm a little confused rn, the indexer is running on our VM and I haven't deployed this yet.

Haroenv:
yes, I applied the setting manually in the dashboard as well, because I wasn't sure when the deploy happened, but I saw the version on GitHub so thought it happened since 😅

MartinKolarik:
Ok, I see, the deploys are manual, so I'll do it now to make sure it doesn't get reverted.

